PR #14, opened 2013-06-22
Extended type support

ramonmaruko:
My C is not really good, so please review this properly. Thank you.


ramonmaruko:
Sorry if it took me too long to implement `::register_extension`. I was busy with other things.

sardaukar:
Great work, thanks!

tagomoris:
My review results are:
* ext type information should be for packer/unpacker instances, not runtime global

Now we should consider about JRuby implementation...

ramonmaruko:
Seems like this isn't needed anymore since #71 is already here.

PR #23, opened 2013-12-03
Visual C++ support.

unak:
* ext/msgpack/sysdep_types.h (bool): Visual C++ in C mode does not have bool.
  define it just like the C99 way.

* others: narrow scoping is a good practice.  but suddly Visual C++ in C mode
  does not support declaring variables except the top of functions.

### DIFF
--- a/ext/msgpack/buffer.c
+++ b/ext/msgpack/buffer.c
@@ -121,6 +121,8 @@ void msgpack_buffer_mark(msgpack_buffer_t* b)
 
 bool _msgpack_buffer_shift_chunk(msgpack_buffer_t* b)
 {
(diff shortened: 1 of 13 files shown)


tagomoris:
What can i do for this?

nurse:
Considered this thread is already done.

cosmo0920:
Visual Studio 2013 has `_Bool` type (also we can write it as `bool` with including stdbool.h).
Perhaps, Visual Studio 2015 solves C99 issue totally.

tagomoris:
I'll close this at once. Please open another issues if there are any additional problems with recent environment.

PR #25, opened 2013-12-14
add encodings support to unpack (compatible with msgpack-jruby)

technorama:


tagomoris:
Now we are merging `msgpack-jruby` into this repository. So, I think that we don't have to take care about compatibility with it.
Please reopen this again if you have any concern about it.

PR #35, opened 2014-08-03
Added packer support for str8, rewrote specs.

CodeMonkeySteve:


tagomoris:
This problem is solved by #62.
Sorry for late reply, and thank you for contribution!

PR #37, opened 2014-08-21
Add support for ActiveSupport's TimeWithZone class

ndbroadbent:
This ensures that TZ data is not stripped. Users will still need to parse the time from the string, but at least it will be serialized correctly.


tagomoris:
It seems uncommon change for msgpack users. Please do it in other gem if you need it.

PR #39, opened 2014-10-02
Fix to use https url instead of :rubygems.

oza:
Signed-off-by: Tsuyoshi Ozawa <ozawa.tsuyoshi@gmail.com>


frsyuki:
yes, this is already done. Thank you

PR #41, opened 2014-10-20
fixed a bug that feed_each returns no data enumerator

kou:
`unpacker.feed_each(data).each` raises the following error:

    ArgumentError: wrong number of arguments (0 for 1)


tagomoris:
See #89

PR #43, opened 2014-11-21
improve error message when you have extra bytes

pmenglund:


okkez:
Looks good to me.
TravisCI says that build error on Ruby1.8.7, but we don't support Ruby1.8.7.

tagomoris:
:+1:

Issue #44, opened 2014-12-03
MsgPack-Ruby library missing support for Binary data type 


mattheworiordan:
I am using MsgPack to encode & decode strings and binary objects across different languages, and I see that the Ruby library makes no attempt to handle binary data according to the Ruby convention of using ASCII-8BIT encoding on byte arrays as strings.

For example:
```ruby
binary_base64 = 'HO4cYSP8LybPYBPZPHQOtuD53yrD3YV3NBoTEYBh4U0N1QXHbtkfsDfTspKeLQFt'
binary_source = Base64.decode64(binary_base64)
binary_source.encoding # => #<Encoding:ASCII-8BIT>

binary_packed = MessagePack.pack(binary_source)
binary_packed.encoding # => #<Encoding:ASCII-8BIT>
binary_unpacked = MessagePack.unpack(binary_packed)

# Now this is where things go wrong, the unpacked encoding is UTF-8, not the original ASCII-8BIT
binary_unpacked.encoding # => #<Encoding:UTF-8>

# Source binary data and unpacked binary data are now not the same
binary_unpacked == binary_source # => false
```

The problem I have is that unless I now attach metadata to the MsgPack payload that indicates whether it is a String (UTF-8) or Binary, my client libraries have no idea how to provide the data back to the consumer of our API i.e. as UTF-8 or ASCII-8BIT. 

@frsyuki, what do you suggest should be done as right now this is unfortunately a blocker for me as I cannot handle binary data for our developer customers?  It would be really appreciated to get your thoughts on this as I need to either work around this issue, attempt to fix this myself, or simply not support binary for our customers at this point which would be disappointing.

PS. The issue https://github.com/msgpack/msgpack-ruby/issues/40 is somewhat related, however it is suggesting all strings are set to ASCII-8BIT by default, whereas I believe the library should detected ASCII-BIT encodings and use the [MsgPack bin format](https://github.com/msgpack/msgpack/blob/master/spec.md#bin-format-family), and for all other encodings it should continue as is and use the [MsgPack string format](https://github.com/msgpack/msgpack/blob/master/spec.md#str-format-family).

jamielennox:
This is catching me as well. I'm sending encrypted binary data within msgpack and the string encoding is messing this up. If i base64 encode then it works however if i was ok with 4x size i wouldn't be using msgpack.

nathany:
@jamielennox Use `gem "msgpack-ably"` for now.

[...]

PR #45, opened 2014-12-03
Bin 8 bit support for Ruby ASCII-8BIT data type

mattheworiordan:
This Pull Request adds support for the [Msgpack bin](https://github.com/msgpack/msgpack/blob/master/spec.md#bin-format-family) data type.  If you refer to http://yehudakatz.com/2010/05/05/ruby-1-9-encodings-a-primer-and-the-solution-for-rails/, you will see that the convention for Ruby binary data is to use the `ASCII-8BIT` encoding.  The Ruby MsgPack gem will now detect `ASCII-8BIT` when encoding and encode as `bin`, and will return `ASCII-8BIT` strings if the data type is bin.  

This PR fixes issue https://github.com/msgpack/msgpack-ruby/issues/44, and also resolves a few small other issues such as adding support for str 8 data type that was missing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 *.bundle
 *.gem
 *.class
+doc
+.yardoc
 .bundle
 Gemfile*
 pkg
--- a/ext/msgpack/packer.h
+++ b/ext/msgpack/packer.h
@@ -271,6 +271,10 @@ static inline void msgpack_packer_write_raw_header(msgpack_packer_t* pk, unsigne
         msgpack_buffer_ensure_writable(PACKER_BUFFER_(pk), 1);
         unsigned char h = 0xa0 | (uint8_t) n;
         msgpack_buffer_write_1(PACKER_BUFFER_(pk), h);
+    } else if(n < 256) {
+        msgpack_buffer_ensure_writable(PACKER_BUFFER_(pk), 2);
+        unsigned char be = (uint8_t) n;
+        msgpack_buffer_write_byte_and_data(PACKER_BUFFER_(pk), 0xd9, (const void*)&be, 1);
     } else if(n < 65536) {
         msgpack_buffer_ensure_writable(PACKER_BUFFER_(pk), 3);
         uint16_t be = _msgpack_be16(n);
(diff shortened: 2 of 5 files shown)


mattheworiordan:
@iconara and @frsyuki is there any reason this isn't going to be merged in?  If not, I will have to create a fork of this gem with a new namespace such as `MessagePack5` so that it does not conflict with this namespace.  It has been well over a month that this PR has been open and it's a real blocker for us as binary support is needed.

iconara:
I'm all for it, but I'm the JRuby guy so I'll have to leave the final say to @frsyuki.

If anything I would have liked to see tests for what happens when a string is not UTF-8 nor ASCII-8BIT. I might be wrong, but it looks like all non-ASCII-8BIT strings are written as raw strings, and from my understanding of the spec that is not right. Raw strings should be UTF-8, so non-UTF-8 strings need to be transcoded before they are packed. Otherwise we're back to the same problem as the old MessagePack spec: you had no idea what encoding strings were in.

Here's how the JRuby implementation handles this case: https://github.com/msgpack/msgpack-ruby/blob/master/ext/java/org/msgpack/jruby/Encoder.java#L194-L196, there's also a (maybe cryptically described) test for it: https://github.com/msgpack/msgpack-ruby/blob/master/spec/jruby/msgpack_spec.rb#L131-L134

mattheworiordan:
Thanks for the quick reply @iconara.
Sure, it wouldn't be difficult to add that support, although equally Binary support is still better than the current state of affairs.  Do you have any idea if @frsyuki will ever consider this change, because if not, I am going to start an new fork as I need MsgPack support in my client library yet if I'm using a fork and the person using my library is already using MsgPack, it will prevent anyone else from my using my library unfortunately.  Any ideas on what I can do?

nathany:
@frsyuki Any feedback on this? I'm happy to use the msgpack-ably fork, but it would be nice if bin was supported directly.

mattheworiordan:
@nathany I am probably going to create a MsgPackV5 fork that uses a different namespace as this fork is not ideal as the namespace would conflict if someone used both the old & forked MsgPack gems.  However, if this is not a problem for you (it won't effect most people) then please do use my fork, I hope it helps.

frsyuki:
Thank you for this pull-request.
As @iconara mentions, transcoding would be necessary if string is encoded in non-UTF8 encodings. Implementation of `is_byte_array` calls rb_eval_string and rb_funcall  but it may impact performance significantly. 
I created #62.

PR #46, opened 2014-12-04
rename msgpack_unpacker_{init,reset,destroy} functions

miihael:
My collegue and me are creating a ruby wrapper for our C library. The library uses https://github.com/msgpack/msgpack-c.git 
We've noticed, that when some program requires both 'msgpack' ruby module and the wrapper, then "Segmentation fault" appears.
The reason - identical names of 3 functions in symbols of msgpack C library and the ruby module.

This patch adds underscores in the names of msgpack_unpacker_{init,reset,destroy} functions. They are used only internally and can't affect any external ruby code.


frsyuki:
:+1:

PR #48, opened 2015-01-16
Add jruby-head to .travis.yml.

SebastianEdwards:
Adds jruby-head to travis config while allowing failures.


tagomoris:
It's already in `.travis.yml`.
Sorry for late reply, and thank you for contribution!

PR #50, opened 2015-02-03
Enable tests in unpack_spec in JRuby

iconara:
I noticed that these tests were skipped in JRuby but they actually pass. @frsyuki do you remember why you disabled them?


frsyuki:
@iconara I'm not sure but it worked. merged.

PR #51, opened 2015-02-05
Support JRuby 9k

grddev:
As JRuby 9k is approaching release, it makes sense to add support for it. Only one tiny code change was required in order to get it up and running (while retaining backwards compatibility). I have verified locally that it works with jruby-head, jruby-1.7.2, and jruby-1.7.6.

I also took the opportunity to add jruby-head to the build matrix (allowing failures) for much the same reason.


iconara:
Thank you!

Issue #42, opened 2014-11-10
Support for Extended types?


alexgenco:
It looks like there isn't support for [extended types](https://github.com/msgpack/msgpack/blob/master/spec.md#extension-type) yet, despite a stale pull request (https://github.com/msgpack/msgpack-ruby/pull/14).

Are there plans to implement it? It's a must-have to integrate with the [Neovim](https://github.com/neovim/neovim) MessagePack API. I would love to help but I've never written C in my life :disappointed: 

Great work with this! :smile:

niamster:
Does it make sense to encode ruby symbols?
In my experience it's a bit annoying when existing RPC server code expects a symbol but unpacked message carries a string. When switching from ruby's native marshalling it's a big effort to check all the places for correctness(disallow sending symbols might be painful as those might be a part of nested arrays and checking everything is just a waste of CPU time).
Is is correct from msgpack point of view encode a type with possible native types(and indicating a fall-back)?

nocelic:
@alexgenco we are working on this, I hope to have a proposal ready in a couple of days. 

@tagomoris moving my question form #52 to #42: 
Do we want the MRI code to match the JRuby code?
The JRuby code uses ExtensionValue class, which I do not really like. I think ExtType (form msgpack-python) or ExtensionValue or Extension or Extended would all have been better choices.
So we can
(a) Mirror the JRuby solution with ExtensionValue in MRI
(b) Change the JRuby code to match the MRI solution, whatever that will be.
(c) Write an optional JRuby compatibility layer in MRI
(d) Have separate interfaces for MRI and JRuby.

Personally, I'd go with (a) in order not to break any possibly existing code.
But before I proceed, I'd be happy to hear your opinion, so that I do not waste my time on something that will be rejected later.

tagomoris:
(a) and/or (c) look good for me.
In sight of compatibility (a) is best, but (c) may be better than (a) if there're any idea to implement faster than (c). Best way to implement ext types may be different between MRI and JRuby.

I seriously need @frsyuki 's comment :smile:

frsyuki:
It is good idea to add a Ruby class that corresponds to MessagePack's Extended type. msgpack-java has [org.msgpack.value.ExtendedValue](https://github.com/msgpack/msgpack-java/blob/v07-develop/msgpack-core/src/main/java/org/msgpack/value/ExtendedValue.java).

[...]

PR #52, opened 2015-02-28
Msgpack custom types

csfrancis:
This is an attempt at resolving #42.

The idea here is to add the ability to serialize custom types using [MessagePack’s extension mechanism](https://github.com/msgpack/msgpack/blob/master/spec.md#types-extension-type). Basic usage is very simple:

```ruby
MessagePack.register_type Time
MessagePack.register_type Date
MessagePack.register_type DateTime
```

By default, the implementation serializes the type using Ruby’s `Marshal`. The upside of this is that it allows for the serialization of any Ruby type, but the downside is that it’s serialized data (Marshal) within serialized data (MessagePack), which has a negative impact on performance.

It’s possible to provide custom serialization for any Ruby type, instead of using `Marshal`. To do this, define a `to_msgpack` instance method on a class which returns a string:
```ruby
class Time
  def to_msgpack
    # pack a Time instance into 12 bytes
    “#{[tv_sec].pack(‘Q>’)}#{[tv_nsec / 1000].pack(‘L>’)}"
  end
end
```

If using custom serialization with `to_msgpack` you must also provide a custom `from_msgpack` class method:
```ruby
class Time
  def self.from_msgpack(data)
    Time.at(*data.unpack(‘Q>L>’))
  end
end
```
If using `to_msgpack` and `from_msgpack` for custom serialization, you must define those methods *before* registering the type with `MessagePack.register_type`.

/cc: @byroot, @fbogsany

### DIFF
--- a/ext/msgpack/custom.c
+++ b/ext/msgpack/custom.c
@@ -26,6 +26,9 @@ static ID s___to_msgpack;
 static ID s_from_msgpack;
 static ID s_dump;
 static ID s_load;
+static ID s_instance_method;
+static ID s_arity;
+
 static VALUE mMarshal;
 static VALUE ePackError;
 
@@ -148,6 +151,12 @@ static VALUE custom_to_msgpack(int argc, VALUE* argv, VALUE self)
     return packer;
 }
 
+static int get_instance_method_arity(VALUE cls, ID method_id)
+{
+    VALUE method = rb_funcall(cls, s_instance_method, 1, ID2SYM(method_id));
+    return FIX2INT(rb_funcall(method, s_arity, 0));
+}
+
 static VALUE MessagePack_register_type_module_method(VALUE self, VALUE cls)
 {
     const char * clsname;
@@ -168,7 +177,7 @@ static VALUE MessagePack_register_type_module_method(VALUE self, VALUE cls)
              * we need to alias it to `__to_msgpack` and it will be called
              * by our `to_msgpack`.
              */
-            if (rb_mod_method_arity(cls, s_to_msgpack) == 0) {
+            if (get_instance_method_arity(cls, s_to_msgpack) == 0) {
                 rb_define_alias(cls, "__to_msgpack", "to_msgpack");
                 define_to_msgpack = has_custom_to_msgpack = true;
             }
@@ -198,6 +207,8 @@ void MessagePack_custom_module_init(VALUE mMessagePack)
     s_from_msgpack = rb_intern("from_msgpack");
     s_dump = rb_intern("dump");
     s_load = rb_intern("load");
+    s_instance_method = rb_intern("instance_method");
+    s_arity = rb_intern("arity");
 
     mMarshal = rb_define_module("Marshal");
 
--- a/ext/msgpack/custom.c
+++ b/ext/msgpack/custom.c
@@ -158,6 +158,8 @@ static VALUE MessagePack_register_type_module_method(VALUE self, VALUE cls)
     if (!st_lookup(msgpack_custom_types, (st_data_t) clsname, NULL)) {
         VALUE singleton_cls = rb_singleton_class(cls);
         bool define_to_msgpack = false;
+        bool has_custom_to_msgpack = false;
+        int has_from_msgpack = 0;
 
         if (!rb_method_boundp(cls, s_to_msgpack, 0)) {
             define_to_msgpack = true;
@@ -168,7 +170,7 @@ static VALUE MessagePack_register_type_module_method(VALUE self, VALUE cls)
              */
             if (rb_mod_method_arity(cls, s_to_msgpack) == 0) {
                 rb_define_alias(cls, "__to_msgpack", "to_msgpack");
-                define_to_msgpack = true;
+                define_to_msgpack = has_custom_to_msgpack = true;
             }
         }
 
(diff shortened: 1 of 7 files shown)


ext/msgpack/custom.c
```diff
@@ -0,0 +1,208 @@
...
+    size_t data_sz;
+    VALUE str, cls, val;
+
+    if (*p++ != MSGPACK_EXT_TYPE_RUBY) {
+        return Qnil;
+    }
+
+    /* only deserialize types that we've registered */
+    if (!st_lookup(msgpack_custom_types, (st_data_t) p, (st_data_t *) &cls)) {
+        return Qnil;
```
byroot: Shouldn't you raise an error?

csfrancis: That will happen, just not from here. By returning `nil`, the unpacker will return `PRIMITIVE_UNEXPECTED_TYPE` which eventually does raise an error: https://github.com/csfrancis/msgpack-ruby/blob/msgpack_custom_types/ext/msgpack/unpacker.c#L348-L353

I didn't want to raise any errors from here to be consistent with the rest of the unpacker code.

byroot: :ok_hand:

csfrancis:
Could I get someone to take a look at this?

/cc: @frsyuki

byroot:
@tagomoris @frsyuki any chance one of you guys could look at this?

tagomoris:
Sorry to have missed it. I'll check this tomorrow.

byroot:
Awesome thanks!

In term of API I wonder if it shouldn't instead use the classic serializer interface:

```ruby
module TimePacker
  def self.load(payload)
    Time.at(*payload.unpack(‘Q>L>’))
  end

  def self.dump(time)
    “#{[tv_sec].pack(‘Q>’)}#{[tv_nsec / 1000].pack(‘L>’)}"
  end
end
```

That would avoid the necessity to monkey patch native types.

csfrancis:
@byroot I kind of like the API the way it is. The extension is already monkey patching the native types: https://github.com/msgpack/msgpack-ruby/blob/03efc747f0382cc9cf9cf501fa247f16777f209f/ext/msgpack/core_ext.c#L118-L127. I think changing the API would require a major version bump.

Also, /cc: @methodmissing

nocelic:
I have read the entire source of this patch and I would NOT recommend merging it into the official msgpack library.  

Why? The way I understand #42, it wants to make the **extended types from the msgpack specification** available to Ruby. This makes #42 a legitimate request, because the current msgpack-ruby is indeed lagging behind the specification due to the lack of support for extended types .

In contrast, this patch seems to serialize **custom types**, i.e. arbitrary classes, and that is **a completely different thing**. Furthermore, this patch seems to hijack the extended type code 1 for that purpose, which is not a part of the specification and should therefore not be a part of the official msgpack library.

IMHO, the proper solution would be to 
1) Implement generic support for extended types from the specification, and then 
2) On top of that, add support for serializing arbitrary classes, preferably in a way that makes it an optional feature.  

As a side note, the msgpack-python has already implemented extended type support in a generic and completely "to-the-specs" way. I would prefer a similarly generic model in Ruby, and not a solution departing from the spec.

tagomoris:
My review results are: 

* this patch will be rejected (conclusion)
* this patch doesn't have any `Extension` interface of MessagePack's spec
* API MUST NOT hide type-information, and users should use type-information explicitly for ext types
* ext type information should be for packer/unpacker instances, not runtime global

(Some of these above are also for #14 )
@nocelic 's comments make sense for me.

Are there any objections? @frsyuki

csfrancis:
Thank you guys for the feedback.

I think the main issue here that I'm adding Ruby specific functionality into the core msgpack gem, which you guys don't like (and I see your point). What I'm going to do is close this, and create another PR that JUST adds Extension interface functionality. The code that I wrote to do custom serialization/deserialization of Ruby objects can go into another gem that leverages the Extension interface.

nocelic:
@tagomoris : 
I just wrote a prototype code that implements the msgpack extension types. I can complete it in a couple more days, but there is one important question: 
Do we want the MRI code to match the JRuby code? 
The JRuby code uses `ExtensionValue` class, which I do not really like. I think `ExtType` (form msgpack-python) or `ExtensionValue` or `Extension` or `Extended` would all have been better choices. 
So we can
(a) Mirror the JRuby solution with `ExtensionValue` in MRI
(b) Change the JRuby code to match the MRI solution, whatever that will be. 
(c) Write an optional JRuby compatibility layer in MRI
(d) Have separate interfaces for MRI and JRuby.

Personally, I'd go with (a) in order not to break any possibly existing code. 
But before I proceed, I'd be happy to hear your opinion, so that I do not waste my time on something that will be rejected later.

nocelic:
@csfrancis : 
Yes, exactly, the lib should first implement a complete specification, then build optional features on top of it. 
Since I have got an implementation of extended types almost ready, you might perhaps want to wait a couple of days to see what comes out of it. 
Alternatively, if you can complete it within a week, please let me know, so that we do not make redundant efforts.

csfrancis:
@nocelic I can wait to take a look at your implementation, please feel free to tag me on a PR.

tagomoris:
@nocelic Could you create an issue for that discussion?
Comments of closed pull-request is not good for history of this library.
Now I'm preparing comment for you for comming issue.

nocelic:
@tagomoris yup, moving the discussion to #42.

PR #54, opened 2015-03-09
Remove unnecesary variables

matugm:


spec/cruby/buffer_spec.rb
```diff
@@ -379,7 +379,6 @@
   end
 
   it 'big write -> short write' do
-    biglen = 1024*1024 + 2
     big1 = "a" * (1024*1024 + 2)
```
cosmo0920: I think above line removing is worthless.  
`big1 = "a" * (1024*1024 + 2)` line should use `big1 = "a" * biglen` for readability instead.

PR #59, opened 2015-05-07
Ensure that JRuby symbolize_keys default is consistent

grddev:
The use of fastARef means that a Java null will be returned in the event that a Hash was provided, but without the symbolize_keys key, which resulted in symbolizeKeys to be turned on even though it wasn't requested.


iconara:
Merged in 8dd92d21de00bab13b52c7f0552e2bf92fb30068

PR #62, opened 2015-05-18
Bin 8 bit support for Ruby ASCII-8BIT data type 2

frsyuki:
This pull-request adds encoding transcoding to #45. If string is not encoded in ASCII-compatible encoding such as UTF-16, or the string includes 7bit unsafe characters, Packer transcoders the string into UTF-8.

### DIFF
--- a/ext/msgpack/unpacker.c
+++ b/ext/msgpack/unpacker.c
@@ -30,6 +30,7 @@ static msgpack_rmem_t s_stack_rmem;
 
 #ifdef COMPAT_HAVE_ENCODING  /* see compat.h*/
 static int s_enc_utf8;
+static int s_enc_ascii_8bit;
 #endif
 
 void msgpack_unpacker_static_init()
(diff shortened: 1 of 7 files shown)


frsyuki:
@tagomoris can I ask you to review this code? I think this code should be released with new major version (v0.6.0) because behavior of Packer changes. It is more than happy to me if you can work on releasing v0.6.0. Thank you.

ext/msgpack/unpacker.c
```diff
@@ -581,7 +585,7 @@ int msgpack_unpacker_read_map_header(msgpack_unpacker_t* uk, uint32_t* result_si
     return 0;
 }
 
-int msgpack_unpacker_read(msgpack_unpacker_t* uk, size_t target_stack_depth)
+    int msgpack_unpacker_read(msgpack_unpacker_t* uk, size_t target_stack_depth)
```
tagomoris: Wrong indentation. (I'll fix it later)

tagomoris:
LGTM, except for fails of tests on JRuby, about raw8/bin8.
But it is written in TODO of `ext/java/org/msgpack/jruby/Encoder.java`, so I'll fix it later.

PR #63, opened 2015-05-18
Fix to pack small str/bin data into str8/bin8, same with #62

tagomoris:
* This fix is actually to reveal comment-outed code of JRuby implementation


tagomoris:
@iconara Could you review this change?

iconara:
If I understand it correctly this is the way it was before the merge with msgpack-ruby, so I think it's good.

tagomoris:
Exactly!
OK, I'll merge this.

PR #64, opened 2015-05-19
Support Visual C++ 12 (Visual Studio 2013)

nurse:
Note that Visual C++ 11 (Visual Studio 2012) doesn't have stdbool.h though it may build msgpack with little effort.


tagomoris:
Merged. Thanks!

Issue #67, opened 2015-06-20
"compatibility mode" for packer


0xBF:
v0.6.0 use new msgpack specification, which add str8 and binary formats. But this formats don't supported by old parsers and generated msgpack data can't be parsed by such clients.

Please, add flag for disable str8 and bin formats in packer.

tagomoris:
@0xBF Are there any reason not to use `v0.5.x`?

0xBF:
how long will be 0.5.x branch supported?
At least Debian Wheezy, Ubutu 14.04 LTS, CentOS 6.x comes with msgpack-c 0.5.7 without new deserialization format.

tagomoris:
`v0.5.x` is available as it is. If coming OS versions breaks anything, it will not works well...
Anyway, patches are welcome for optional packer feature.

iconara:
This is what the [MessagePack specification](https://github.com/msgpack/msgpack/blob/master/spec.md#impl-upgrade) has to say on the subject:

> At the same time, serializers should offer "compatibility mode" which doesn't use bin format family and str 8 format

I will take a stab at fixing this, because I need it too. I might need help with the C implementation.

tagomoris:
#68 and #70 are merged, and `v0.6.1` released with option `compatibility_mode: true` for Packer.
That solves this issue.

PR #68, opened 2015-06-30
Compatibility mode

iconara:
This adds an option to `MessagePack.pack` and `MessagePack::Packer.new` that makes the packer work in what the [MessagePack specification calls "compatibility mode"](https://github.com/msgpack/msgpack/blob/master/spec.md#upgrading-messagepack-specification) – it disables the bin and str8 types.

This should solve #67.

Please note that I know next to nothing about writing Ruby C extensions. Someone, perhaps @tagomoris, should look at that and give me hints to how to do it properly.

I'm not super happy about calling the option `:compatibility_mode`, so if there are any other suggestions I would be happy to change it.

### DIFF
--- a/ext/java/org/msgpack/jruby/Buffer.java
+++ b/ext/java/org/msgpack/jruby/Buffer.java
@@ -21,7 +21,6 @@ import org.jcodings.Encoding;
 
 @JRubyClass(name="MessagePack::Buffer")
 public class Buffer extends RubyObject {
-  private RubyHash options;
   private IRubyObject io;
   private ByteBuffer buffer;
   private boolean writeMode;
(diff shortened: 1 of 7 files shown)


tagomoris:
How about `compat` for option name? The word `compat` is used for such cases historically.
And now I'm checking code.

iconara:
I must say I don't see any benefit in using the shorter form "compat" over "compatibility mode". I was thinking of something that describes it better, that explains _what is it compatible with_. I want to express "backwards compatible" and "don't use new features" and "raw not bin or str8".

tagomoris:
We don't have appropriate name for what it is compatible with...
IMO just saying `compat` means what is compatible with behavior of older versions. But I agree with your comment: explaining longer name is better than shorter one.

tagomoris:
@iconara I added complete cruby implementation. please merge #68.

iconara:
Merged to master.

frsyuki:
:+1:

tagomoris:
#70 (instead of #69) also merged. I'll release v0.6.1.

iconara:
That's great, thanks for the review and the cruby code.

PR #69, opened 2015-07-01
Add "compatibility_mode" option for CRuby implementation

tagomoris:
Add commits for cruby on #68


tagomoris:
I'll use #70 instead of this pull request.

PR #71, opened 2015-07-02
Add support for extension type and application-defined class serializers

frsyuki:
working in progress.


tagomoris:
It seems required to check type/Class already exists (and raise ArgumentError or something) for `register_type`.

tagomoris:
Just idea: how do you think about that `MessagePack.pack` and `MessagePack.unpack` refer `MessagePack::DefaultFactory`?

tagomoris:
`MessagePack::Factory.new.register_type(type_id, Klass)` successes (it uses `:to_msgpack_ext` and `:from_msgpack_ext`).
But these fails:
* `MessagePack::Packer.new.register_type(type_id, Klass) # no block given (LocalJumpError)`
* `MessagePack::Unpacker.new.register_type(type_id, Klass) # wrong number of arguments (2 for 1 or 3) (ArgumentError)`

@frsyuki How do you think about to check `Klass#to_msgpack_ext` and `Klass.from_msgpack_ext` and to use it if it exist?

tagomoris:
`MessagePack::ExtensionValue` should have instance method `#to_msgpack`, right?
(JRuby implementation has it).

frsyuki:
> a) check type/Class already exists

Just allowing users to overwrite type/Class should be OK as you mention.
Rather I considered that `#register_type` returns former Proc if one is set. But I also stopped doing it because return value is not clear if `#register_type` is called with method name in Symbol.

> b) MessagePack.pack and MessagePack.unpack refer MessagePack::DefaultFactory

It's good idea. When I was writing test cases, I wanted `Factory#pack` and `Factory#unpack` An idea is to assume that `MessagePack.pack` is an alias of `MessagePack::DefaultFactory.pack` and same for `.unpack`. Method name (`factory.pack(obj)`) is weird but useful. How do you think?

> c) `Packer#register_type(type_id, Klass)`, `Unpacker#register_type(type_id, Klass)`

I agree the basic idea but checking existence of an instance method in Ruby is a bit tricky because of singleton methods and method_missing.
For now, `Factory#register_type` does not check existence of `Klass#to_msgpack_ext` method because it simply calls `:to_msgpack_ext.to_proc()`.
It is easy to use `Klass#to_msgpack_ext` if block is not given to `Packer#register_type(type_id, Klass)`. Any ideas?

frsyuki:
hm, for c), my idea is to check existence of method using `Klass.instance_method(:to_msgpack_ext)` and ignore the returned value. The purpose is validation to provide fail-fast policy.

tagomoris:
@frsyuki 

> a) check type/Class already exists

If overwriting definitions are permitted, we must have an API to check definitions exist or not or to query types or classes, to make users enable to check wether they will overwrite existing definition or not before `register_type`.
Return values (after overwriting definitions) seems not good idea for me.

> b) MessagePack.pack and MessagePack.unpack refer MessagePack::DefaultFactory

`factory.pack(obj)` seems something weird for me too. It's better to undocument about `MessagePack::Factory#pack and #unpack` if we take that approach.

But I think that it's more better just to use `MessagePack.pack` and `MessagePack.unpack` directly.
Are there any problem about to use packer/unpacker from MessagePack::DefaultFactory in MessagePack.pack and MessagePack.unpack?

> c) `Packer#register_type(type_id, Klass)`, `Unpacker#register_type(type_id, Klass)`

In my idea, checks are not needed at registered time. We can extend class definitions anytime after registration into packers/unpackers in ruby way (just same with defining `#to_msgpack` and calling it).

frsyuki:
I noticed that a callback registered by `Packer#register_type(type, klass)` needs to match an object which **extends** `klass` as well as class == `klass`. So, I guess that `Packer` needs to use `Class#===` to lookup a registered callback. In terms of implementation, `msgpack_packer_ext_registry_lookup` uses `rb_hash_lookup` currently but it needs to be changed to use linear search + `rb_funcall(registered_class, rb_intern("==="), 1, obj)`.

* a) check type/Class already exists

I see. We need to think about the semantics. How do you think about this idea?

1. Add `Unpacker#type_registered?(type_or_class)`. This method returns true if
    * type_or_class.is_a?(Integer) && type_or_class == ext_type of a registered entry, or
    * !type_or_class.is_a?(Integer) && type_or_class **<=** class of a registered entry
2. Add `Packer#type_registered?(type_or_class)`. This method returns true if
    * type_or_class.is_a?(Integer) && type_or_class == ext_type of a registered entry, or
    * !type_or_class.is_a?(Integer) && type_or_class **<=** class of a registered entry
3. Add `Unpacker#type_registered?(args...)` (overload). This method returns true if
    * `#register_type` was called before with the same `args...`.
4. Add `Packer#type_registered?(args...)` (overload). This method returns true if
    * `#register_type` was called before with the same `args...`.

* b) MessagePack.pack and MessagePack.unpack refer MessagePack::DefaultFactory

I reimplemented `MessagePack.pack` and `MessagePack.unpack` using `MessagePack::DefaultFactory`. I think this is just useful without problems.
We can add `Factory#pack` and `Factory#unpack` when necessary.

* c) Packer#register_type(type_id, Klass), Unpacker#register_type(type_id, Klass)

Sounds good. It's dynamically typed way.

tagomoris:
@frsyuki Seems good for b) and c).

For a), I will:
* implement `#registered_types` for `Packer`, `Unpacker` and `Factory` which return Hash of mapping in native code
* implement utility methods like `type_registered?(type_or_class)` in ruby code

IMO, this approach makes us to provide many utility methods, to share implementations between ruby and jruby and makes users familiar to propose a new pull-request for another API :smile:

tagomoris:
Added some fixes for Rubinius, and marks of `skip` for JRuby environment.
And then rebased on current master HEAD.

tagomoris:
I've added some missing specs and doclib contents.
And I also confirmed that specs runs successfully in both of cruby and jruby (which has many skipped specs, though).

@frsyuki I'm going to merge this pull-request. How do you think about it?

frsyuki:
@tagomoris Thank you for the productization work! Looks totally good. Can I ask you merge and release?

tagomoris:
Merged! Yay!
#87 is following issue.

PR #72, opened 2015-07-07
Add methods `#registered_types` and `#type_registered?`

tagomoris:
on Packer, Unpacker and Factory
This is working branch on the top of #71

### DIFF
--- a/ext/msgpack/factory_class.c
+++ b/ext/msgpack/factory_class.c
@@ -182,9 +182,9 @@ static VALUE Factory_register_type(int argc, VALUE* argv, VALUE self)
         }
     }
 
-    msgpack_packer_ext_registry_put(&fc->pkrg, ext_class, ext_type, packer_proc);
+    msgpack_packer_ext_registry_put(&fc->pkrg, ext_class, ext_type, packer_proc, packer_arg);
 
-    msgpack_unpacker_ext_registry_put(&fc->ukrg, ext_type, unpacker_proc);
+    msgpack_unpacker_ext_registry_put(&fc->ukrg, ext_class, ext_type, unpacker_proc, unpacker_arg);
 
     return Qnil;
 }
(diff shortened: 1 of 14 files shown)


tagomoris:
@frsyuki Can you review this change?
Please merge into `register_type_wip` if ok. Then, I'll write more tests for packer/unpacker/factory and update API documents on that branch.

tagomoris:
Pushed one more commit to show how use registered for ext when they registered.
@frsyuki Please take a look at this change.

ext/msgpack/factory_class.c
```diff
@@ -187,6 +200,7 @@ void MessagePack_Factory_module_init(VALUE mMessagePack)
     rb_define_method(cMessagePack_Factory, "packer", MessagePack_Factory_packer, -1);
     rb_define_method(cMessagePack_Factory, "unpacker", MessagePack_Factory_unpacker, -1);
 
+    rb_define_method(cMessagePack_Factory, "registered_types_internal", Factory_registered_types_internal, 0);
```
frsyuki: IMO `rb_define_private_method` is better.

frsyuki:
Looks good :+1: 
Caching result of `msgpack_packer_ext_registry_lookup` sounds good idea. Great job.
Potential problems are:

* If class definition is changed after caching, msgpack_packer_ext_registry_lookup returns unexpected result.
* If application defines classes dynamically, size of cache increases infinitely

But I think those cases are rare. We can ignore them at least for now.

ext/msgpack/factory_class.c
tagomoris: Exactly.

tagomoris:
I understand your concern, but we have a way to clear cache to enforce users to call `register_type` if they make any changes on class definitions. It should be written in documents.

PR #75, opened 2015-07-17
Segfault fixed

beelyl0:
It fixed a segfault related by incorrect work with strings


frsyuki:
Thank you!!! Does rb_id2str return 0...?

frsyuki:
@beelyl0 I asked @nurse why rb_id2str returns 0. We need some investigation also in Ruby.
Could you tell me which version of ruby you used?

If possible, could you give us a code to reproduce segfault at msgpack_packer_write_symbol_value?

PR #79, opened 2015-07-22
Replace project-proprietary Mingw build solution by rake-compiler-dock.

larskanis:
[Rake-compiler-dock](https://github.com/rake-compiler/rake-compiler-dock) is a more common build tool and environment for Windows RubyInstaller compatible binary gems. It supports ruby 1.8 to 2.2 and Windows, Linux and OS-X as the host OS. It runs on top of docker as well.

This fixes https://github.com/msgpack/msgpack-ruby/issues/73

This also updates rake-compiler to the latest version.

I tested the build process on Windows and Linux and I tested the resulting gems on Windows-7 with ruby-1.9.3-x86, ruby-2.0.0-x86, ruby-2.2.2-x86 and ruby-2.2.2-x64, so far.

PR #81, opened 2015-07-23
Add appveyor.yml for ci-tests on Windows.

larskanis:
The github project needs to be enabled in appveyor, similar to travis-ci. The build and specs runs green, like [here](https://ci.appveyor.com/project/larskanis/msgpack-ruby/build/1.0.3) .

appveyor is slower than travis-ci. That's why I only enabled the most opposed ruby versions, only, for now.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,18 @@
+---
+install:
+  - SET PATH=C:\Ruby%ruby_version%\bin;%PATH%
+  - ruby --version
+  - gem --version
+  - bundle install
+build: off
+test_script:
+  - bundle exec rake -rdevkit
+
+environment:
+  matrix:
+    - ruby_version: "200"
+    - ruby_version: "200-x64"
+    - ruby_version: "21"
+    - ruby_version: "21-x64"
+    - ruby_version: "22"
+    - ruby_version: "22-x64"


tagomoris:
Ruby 1.9.3 is not supported by now.
Please remove it from yml file.

larskanis:
With PR https://github.com/msgpack/msgpack-ruby/pull/79 Ruby-1.9.3 could be supported on Windows. It seems to be supported on other platforms as well.

cosmo0920:
Ruby 1.9.3 is already reached EOL. https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/
It should not be support on Windows.
And Ruby 2.0.0 is reached the state of the security maintenance phase.
We should recommend to use Ruby 2.1 or later, I think.

larskanis:
I updated this PR and removed Ruby-1.9.3. Some versions fail, but it can be fixed as in #79.

As a notice: Unfortunately many gems are still not yet prepared for Ruby-2.2 on Windows. This often forces to stay longer on old versions of Ruby on Windows than on other platforms. IMHO gemspec should set a requirement for the ruby version, if the Windows binary gems don't support ruby < 2.0.

tagomoris:
Thanks!

Issue #85, opened 2015-08-03
Make exception behavior of Unpacker consistent to throw only UnpackerError


frsyuki:
As #83 points out, Unpacker throws TypeError, which doesn't extend UnpackerError.
It would be confusing if Unpacker throws an error which is not UnpackerError in terms of its naming.
Possible solutions are...:

* a) change TypeError to extend UnpackerError
* b) change TypeError to be Module. And create UnexpectedTypeError which includes TypeError and extends UnpackerError.

tagomoris:
#86 solved this issue.

PR #86, opened 2015-08-03
Unpacker is expected to raise UnpackeError

### LINKED ISSUE #83: Why TypeError is not subclass of UnpackError?
It is strange, that MessagePack::TypeError is a separate species. Why it is not sublclass of UnpackError?

frsyuki:
This change adds `MessagePack::UnexpectedTypeError` which extends `MessagePack::UnpackError`. Unpacker raises it instead of TypeError.

This doesn't take alternative idea which changes TypeError to extends UnpackError for 2 reasons:

* For backward compatibility with existent code which expects that rescuing TypeError doesn't rescue UnpackError.
* Name of `TypeError` is confusing with Ruby's standard `::TypeError`.

For backward compatibility, TypeError is changed to Module and UnexpectedTypeError includes it.

Fixes #83 and #85.

### DIFF
--- a/doclib/msgpack/error.rb
+++ b/doclib/msgpack/error.rb
@@ -9,9 +9,6 @@ module MessagePack
   class StackError < UnpackError
   end
 
-  module TypeError
-  end
-
   class UnexpectedTypeError < UnpackError
     include TypeError
   end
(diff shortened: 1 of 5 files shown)


tagomoris:
@frsyuki Code looks good to me.
Could you rebase once? (I've merged a big pull-request what you know :)

frsyuki:
rebased: https://github.com/msgpack/msgpack-ruby/pull/88

frsyuki:
Rebased. I added UnknownExtTypeError error class which is used only when `allow_unknown_ext` option is false (default behavior).

tagomoris:
OK, LGTM. Merged!

PR #89, opened 2015-08-27
fixed a bug that feed_each returns no data enumerator (revised)

cosmo0920:
Revised #41.
I resolved conflict and rebased with latest master.

`unpacker.feed_each(data).each` raises the following error:

    ArgumentError: wrong number of arguments (0 for 1)


tagomoris:
Great fix. Thanks!

Issue #30, opened 2014-03-06
NoMethodError: undefined method `to_msgpack' for 2014-03-06 15:46:12 +0000:Time


hackeron:
I'm trying to message pack a hash with a time in it but I'm getting this:

```ruby
[9] pry(main)> { a: Time.now }.to_msgpack
NoMethodError: undefined method `to_msgpack' for 2014-03-06 15:47:18 +0000:Time
from (pry):10:in `to_msgpack'
```

How do I handle time with msgpack?

tagomoris:
Time serialization/deserialization will be supported by ext types of version `v0.7.x`.
I'll leave this issue to be opened until that release.

tagomoris:
`v0.7.0` is now available.

ThomasKoppensteiner:
How does the "support by ext types" work?
For me this is still an issue (on JRuby-9.1.15.0 and msgpack 1.2.2):

```
> require "Date"
=> true
> require "msgpack"
=> true
> d = Date.today
=> #<Date: 2018-02-02 ((2458152j,0s,0n),+0s,2299161j)>
> t = Time.now
=> 2018-02-02 13:24:33 +0100
> d.to_msgpack
NoMethodError: undefined method `to_msgpack' for #<Date: 2018-02-02 ((2458152j,0s,0n),+0s,2299161j)>
> t.to_msgpack
NoMethodError: undefined method `to_msgpack' for 2018-02-02 13:24:33 +0100:Time
> d.to_s.to_msgpack
=> "\xAA2018-02-02"
> t.to_s.to_msgpack
=> "\xB92018-02-02 13:24:33 +0100"
```

Issue #19, opened 2013-08-22
Date/Time Not actually in UTC


skizot722:
During packing, datetime/time fields are not actually being converted to UTC.

1.9.3p429 :099 > MessagePack.unpack(MessagePack.pack(MyThing.find(38432).serializable_hash))['created_at']
  MyThing Load (0.9ms)  SELECT "my_things".* FROM "my_things" WHERE "my_things"."id" = $1 LIMIT 1  [["id", 38432]]
 => "Time[2013-08-22 16:59:20 UTC]" 

1.9.3p429 :100 > MyThing.find(38432).created_at
  MyThing Load (0.9ms)  SELECT "my_things".* FROM "my_things" WHERE "my_things"."id" = $1 LIMIT 1  [["id", 38432]]
 => Thu, 22 Aug 2013 16:59:20 CDT -05:00 

Table "public.my_things"
 id           | integer                     | not null default nextval('my_things_id_seq'::regclass)
 created_at   | timestamp without time zone | not null
 updated_at   | timestamp without time zone | not null

The first command that pack's and unpack's should show "2013-08-22 21:59:20 UTC", yet it shows "2013-08-22 16:59:20 UTC".  When packing, it should be calling .utc() on the date, and it's not.  Any ideas?

cosmo0920:
Time serialization/deserialization is not supported current msgpack.
This feature will be supported by ext types of version v0.7.x.

Related to PR #90 discussions.

tagomoris:
Now msgpack-ruby v0.7 supports ext types. So you can serialize Time (or Date) by code like below:
```ruby
MessagePack::DefaultFactory.register_type(0x01, Time, packer: ->(t){ t.to_i.to_msgpack }, unpacker: ->(d){ Time.at(MessagePack.unpack(d)) })
t = Time.now
packed_time = MessagePack.pack(t)
MessagePack.unpack(packed_time) #=> 2015-11-08 13:33:59 +0900
```

PR #90, opened 2015-08-27
Add support for ActiveSupport's TimeWithZone class

okkez:
So that TZ data is not stripped

Original author is Nathan Broadbent <nathan.f77@gmail.com>.

See also #37

### DIFF
--- a/lib/msgpack.rb
+++ b/lib/msgpack.rb
@@ -1,4 +1,5 @@
 require "msgpack/version"
+require "msgpack/active_support/time_with_zone"
 
 if defined?(RUBY_ENGINE) && RUBY_ENGINE == "jruby" # This is same with `/java/ =~ RUBY_VERSION`
   require "java"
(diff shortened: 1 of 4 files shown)


okkez:
Rebase and fix spec.

cosmo0920:
:+1:

frsyuki:
Time doesn't implement to_msgpack. Why only ActiveSupport::TimeWithZone implements to_msgpack?

okkez:
Thank you for reply.
I'm not familiar with msgpack. (On the other hand I'm familiar with Ruby)

But PR #37 is left for a long time and outdated.
So I have updated the patch.

I think that this feature is optional. This patch do nothing when user doesn't have ActiveSupport.
So I think we can merge this and provide optional functionality for convenient.

You can close/reject #37 and this PR if you don't want to merge this feature.
See also #19 and #30 

Thanks.

tagomoris:
It seems uncommon change for msgpack users. Please do it in other gem if you need it.

Issue #87, opened 2015-08-05
Add support `register_type` and `MessagePack::Factory` in JRuby


tagomoris:
Pathes are needed to follow CRuby implementation of these features:
* `MessagePack::Factory`
* `register_type`, `registered_types` and `type_registered?` for `Packer`, `Unpacker` and `Factory`
* Many skipped specs
* Many specs to be shared with CRuby/JRuby (currently many specs are separated on cruby/jruby directory)

tagomoris:
#91 will solve this issue.

Issue #55, opened 2015-04-12
JRuby's ExtensionValue, can't access `type` and `payload`


ramonmaruko:
Is this intended? How do I access the `type` and `payload` properly in Ruby code?

cc @iconara

iconara:
If I remember correctly it's not a finished feature. The JRuby version of msgpack-ruby was in a transition from the old API and old MessagePack standard to the API msgpack-ruby has now and the new MessagePack standard when it got merged in, and I haven't had any time to complete that transition.

If you're using MessagePack extensions and have the time we would be happy to see a PR that adds accessors for `type` and `payload` to `ExtensionValue`. It wouldn't hurt with a few tests around encoding and decoding extensions too.

ramonmaruko:
I'll try to add that to JRuby when I have free time. I already submitted a pull request #14 with a few tests; but only for MRI for now.

iconara:
That would be great, thank you.

tagomoris:
#84 already solved this issue.

PR #91, opened 2015-09-07
add full spec ext type support to JRuby implementation

tagomoris:
This WIP branch is for #87 and #55.
msgpack-ruby v0.7.0 will be released after this branch merged.


tagomoris:
Almost implementation is already done.
Now I'm merging specs between cruby and jruby, but some of these are failing.

tagomoris:
Mission accomplished! Almost all specs are merged into `spec/*.rb`, except for Buffer APIs of CRuby implementation, or `Unpacker#execute` of JRuby implementation. These are left on `spec/cruby` and `spec/jruby`.

Tests are failing with jruby-19mode on mac environment because of Travis-CI's any troubles, and with jruby-head for unidentified error. 

I changed the behavior of JRuby implementation below:
* removed `Unpacker.new(encoding: "ENCODING_NAME")` to decode strings as utf8, and binaries as ascii-8bit
* added `Unpacker.new(allow_unknown_ext: bool)` to specify to allow using Extension objects for unknown ext types, and make that default as false (changed from v0.6 jruby implementation behavior)

@iconara How do you think about this change? Could you review this?

ext/java/org/msgpack/jruby/Decoder.java
```diff
@@ -118,7 +140,20 @@ private IRubyObject consumeHash(int size) {
   private IRubyObject consumeExtension(int size) {
     int type = buffer.get();
     byte[] payload = readBytes(size);
-    return ExtensionValue.newExtensionValue(runtime, type, payload);
+
+    if (this.registry != null) {
+      IRubyObject proc = this.registry.lookup(type);
```
iconara: You don't need to prefix all instance variable accesses with `this.`. It's not that it matters if you do, just that it breaks the style of the rest of the Java code.

ext/java/org/msgpack/jruby/Encoder.java
```diff
@@ -338,6 +338,29 @@ private void appendExtensionValue(ExtensionValue object) {
     buffer.put(payloadBytes.unsafeBytes(), payloadBytes.begin(), payloadSize);
   }
 
+  private void appendExtensionValue(ExtensionValue object) {
+    long type = ((RubyFixnum)object.get_type()).getLongValue();
+    if (type < -128 || type > 127) {
+	    throw object.getRuntime().newRangeError(String.format("integer %d too big to convert to `signed char'", type));
```
iconara: This line is indented with both tabs and spaces.

ext/java/org/msgpack/jruby/Encoder.java
```diff
@@ -338,6 +338,29 @@ private void appendExtensionValue(ExtensionValue object) {
...
+    if (this.registry != null) {
+      IRubyObject[] pair = this.registry.lookup(object.getType());
+      if (pair != null) {
+        RubyString bytes = pair[0].callMethod(runtime.getCurrentContext(), "call", object).asString();
+        int type = (int) ((RubyFixnum) pair[1]).getLongValue();
+        appendExt(type, bytes.getByteList());
+        return;
+      }
+    }
+    // registry is null or type is not registered
```
iconara: I don't like comments in code, they are usually incorrect and no one ever removes them when the code changes. It's better to just not have them in the first place. If you feel that the code doesn't explain what the comment says then wrap the code in a method with a name that explains what's going on.

ext/java/org/msgpack/jruby/Factory.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  @JRubyMethod(name = "unpacker", optional = 1)
+  public Unpacker unpacker(ThreadContext ctx, IRubyObject[] args) {
+    return Unpacker.newUnpacker(ctx, unpackerRegistry(), args);
+  }
+
+  @JRubyMethod(name = "registered_types_internal", visibility = PRIVATE)
+  public IRubyObject registeredTypesInternal(ThreadContext ctx) {
+    // unpacker
```
iconara: This commend doesn't explain anything, it only confuses.

ext/java/org/msgpack/jruby/Factory.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    IRubyObject[] ary = { packerRegistry().hash, mapping };
+    return RubyArray.newArray(ctx.getRuntime(), ary);
+  }
+
+  @JRubyMethod(name = "register_type", required = 2, optional = 1)
+  public IRubyObject registerType(ThreadContext ctx, IRubyObject[] args) {
+    // register_type(type, Class)
+    // register_type(type, Class, packer: proc-like, unpacker: proc-like)
```
iconara: I assume these comments are some kind of type declaration to show which calls can be made, but that should be API documentation, or even better shown with tests.

ext/msgpack/packer_ext_registry.c
```diff
@@ -78,6 +78,7 @@ VALUE msgpack_packer_ext_registry_put(msgpack_packer_ext_registry_t* pkrg,
     return rb_hash_aset(pkrg->hash, ext_class, e);
 }
 
+// TODO: delete this function (unused)
```
iconara: Either delete the function or delete this comment.

ext/msgpack/packer_ext_registry.h
```diff
@@ -95,6 +95,7 @@ static inline VALUE msgpack_packer_ext_registry_lookup(msgpack_packer_ext_regist
     return Qnil;
 }
 
+// TODO: delete this function (unused)
```
iconara: Either delete the function or delete this comment.

ext/msgpack/unpacker_ext_registry.c
```diff
@@ -61,6 +61,7 @@ VALUE msgpack_unpacker_ext_registry_put(msgpack_unpacker_ext_registry_t* ukrg,
     return before;
 }
 
+// TODO: delete this function (unused)
```
iconara: Either delete the function or delete this comment.

ext/msgpack/unpacker_ext_registry.h
```diff
@@ -56,6 +56,7 @@ static inline VALUE msgpack_unpacker_ext_registry_lookup(msgpack_unpacker_ext_re
     return rb_ary_entry(e, 1);
 }
 
+// TODO: delete this function (unused)
```
iconara: Either delete the function or delete this comment.

spec/jruby/unpacker_spec.rb
```diff
@@ -229,15 +138,15 @@ def flatten(struct, results = [])
       Encoding.default_external = Encoding::ISO_8859_1
     end
 
-    it 'produces results with default internal encoding' do
+    it 'produces results with encoding as binary or string(utf8)' do
       unpacker.execute(buffer, 0)
       strings = flatten(unpacker.data).grep(String)
-      strings.map(&:encoding).uniq.should == [Encoding.default_internal]
+      strings.map(&:encoding).uniq.sort{|a,b| a.to_s <=> b.to_s}.should == [Encoding::ASCII_8BIT, Encoding::UTF_8]
```
iconara: I think `strings.map(&:encoding).uniq.should =~ [Encoding::ASCII_8BIT, Encoding::UTF_8]` does the same thing (i.e. contains only, but order independent).

spec/unpacker_spec.rb
```diff
@@ -1,6 +1,257 @@
...
 describe MessagePack::Unpacker do
+  let :unpacker do
+    MessagePack::Unpacker.new
+  end
+
+  let :packer do
+    MessagePack::Packer.new
+  end
+
+  # TODO initialize
```
iconara: Do what the TODO says or remove the comment. Don't merge with TODOs in the code.

iconara:
It's great that the tests are (mostly) unified now, good work. I've done a first pass through the code and commented on some superficial things. I will try to get some more time to understand the structure of how the extension support is implemented.

I commented a lot on code comments. I don't like them, and if you disagree then feel free to ignore what I say about them. I think that comments are either unnecessary because they only repeat what the code does, or they are a sign that the code hasn't been written clearly enough to explain what it does, or it's just noise. Either way, writing the code in such a way that it clearly shows what's going on is always better. Moving a block of code into a method and giving the method a good name is always better than a comment. Don't add "TODO" style comments, do the TODO. If you don't, then it's probably not important, and no one else will do it either. Don't add type signature comments, add API documentation or even better: add tests that show how the method is used.

iconara:
You've checked in a file called `ext/msgpack/.factory_class.c.swp` which probably shouldn't be there, since you also added `*.swp` to the ignores.

iconara:
There's a `eval("return") if java?` line at the top of `spec/factory_spec.rb` so those tests don't run in JRuby. Removing that line gives a few test cases that crash with null pointer and array index out of bounds exceptions.

ext/java/org/msgpack/jruby/Packer.java
```diff
@@ -30,20 +35,139 @@ public IRubyObject allocate(Ruby runtime, RubyClass type) {
     }
   }
 
+  static class ExtRegistry {
```
iconara: Call this `ExtensionRegistry`. Short names are hard to understand; long, descriptive names don't cost anything.

iconara: The same obviously applies to the other ExtRegistry.

iconara:
I don't really see the point in having two separate ExtRegistry implementations. Sure, the packer and unpacker use it differently, but it just feels like having one implementation would be so much simpler to maintain.

ext/java/org/msgpack/jruby/Packer.java
```diff
@@ -30,20 +35,139 @@ public IRubyObject allocate(Ruby runtime, RubyClass type) {
...
+    private Ruby runtime;
+    public RubyHash hash;
+    public RubyHash cache;
```
iconara: I think the code below gets really complicated because it uses `RubyHash` where a Java `Map` would work better. The only reason to use `RubyHash` is when you need to return it to Ruby code, but it doesn't look like that happens here.

ext/java/org/msgpack/jruby/Packer.java
```diff
@@ -30,20 +35,139 @@ public IRubyObject allocate(Ruby runtime, RubyClass type) {
...
+
+    public ExtRegistry dup() {
+      ExtRegistry copy = new ExtRegistry(this.runtime);
+      copy.hash = (RubyHash) this.hash.dup(runtime.getCurrentContext());
+      copy.cache = RubyHash.newHash(runtime);
+      return copy;
+    }
+
+    public void put(RubyClass klass, int typeId, IRubyObject proc, IRubyObject arg) {
+      RubyArray e = RubyArray.newArray(runtime, new IRubyObject[] { RubyFixnum.newFixnum(runtime, typeId), proc, arg });
```
iconara: The code in the `lookup` method gets really complicated to follow because you're using arrays as values. Add a `Entry` class with three public instance variables instead. That would mean that the code in `lookup` would use `entry.typeId` instead of `e.entry(0)`.

Again, using `RubyHash` and `RubyArray` gets really messy in Java code. Use them only when you need to return the objects to Ruby code.

ext/java/org/msgpack/jruby/Packer.java
iconara: Actually, why are these even public? They aren't accessed (and should absolutely not be)

ext/java/org/msgpack/jruby/Packer.java
```diff
@@ -30,20 +35,139 @@ public IRubyObject allocate(Ruby runtime, RubyClass type) {
...
     this.buffer.initialize(ctx, args);
+    this.registry = new ExtRegistry(ctx.getRuntime());
     return this;
   }
 
-  @JRubyMethod(name = "write")
+  public void setExtRegistry(ExtRegistry registry) {
+    this.registry = registry;
+    this.encoder.setRegistry(registry);
+  }
```
iconara: This method should not be needed, the registry should be injected in the constructor.

ext/java/org/msgpack/jruby/Packer.java
tagomoris: These are used by `registered_types_internal` method for `#registered_types` and `#type_registered?` methods of `Factory`, `Packer` and `Unpacker` defined under `lib/msgpack`.
Using RubyHash/RubyArray is just for implementation efficiency. Of course we can convert Java objects into RubyHash/RubyArrah in `registered_types_internal` if you want to do it when refactoring.

ext/java/org/msgpack/jruby/Packer.java
tagomoris: Extension registries are internal object, so it shouldn't visible as argument of `initialize`, and it should be set by java method, not ruby method (`setExtentionRegistry`).

tagomoris:
I've pushed commits against reviews.
JRuby 9.0.1.0 is still failing for tests about encoding...

tagomoris:
Failure on JRuby 9.0.1.0 is from this bug: https://github.com/jruby/jruby/issues/3405

spec/jruby/unpacker_spec.rb
tagomoris: IMO `should =~ [array]` is highly rspec-ish operator, and there's no strong reason to use it.

tagomoris:
@iconara Could you review this change once more?

I didn't change `ExtensionRegistry` internal classes. These are similar but different from each other.
Of course we can create a registry classes for both of Packer and Unpacker, but it will have 2 different member and accessor for Packer and Unpacker. It looks nonsense for me.

spec/jruby/unpacker_spec.rb
iconara: Since this is RSpec that feels like a strange argument. Why not use the tool you're using?

ext/java/org/msgpack/jruby/Decoder.java
```diff
@@ -56,6 +74,10 @@ public void symbolizeKeys(boolean symbolize) {
     this.symbolizeKeys = symbolize;
   }
 
+  public void allowUnknownExt(boolean allow) {
+    this.allowUnknownExt = allow;
+  }
+
```
iconara: This should be a constructor argument (same goes for `#symbolizeKeys`).

ext/java/org/msgpack/jruby/Encoder.java
```diff
@@ -45,6 +46,14 @@ public Encoder(Ruby runtime, boolean compatibilityMode) {
     this.compatibilityMode = compatibilityMode;
   }
 
+  public void setRegistry(Packer.ExtensionRegistry registry) {
+    this.registry = registry;
+  }
```
iconara: This should be a constructor argument.

tagomoris: It can't be, because registries will be overwritten when initialized from factories.
https://github.com/msgpack/msgpack-ruby/blob/ext-type-jruby/ext/java/org/msgpack/jruby/Packer.java#L114

ext/java/org/msgpack/jruby/Decoder.java
tagomoris: Fixed with review, and pushing

iconara:
I think it's harder to explain than show, so I'm going to push up some code in a separate branch showing you what I mean. I'll try to get it done within the next hour or two.

iconara:
My opinion on how this should work is now in #92

tagomoris:
@iconara Thank you! I'm checking it.

tagomoris:
@iconara Your code looks fine for me too.
And I've added just few changes to return compatible objects for `registered_types_internal` to make tests to pass. It doesn't requires so large patches.

I'll merge this branch into master, and then execute re-indentation with 4 spaces for 1 indent level (same with many Java code, like JRuby). Is it ok for you?

tagomoris:
Anyway, merged!

PR #93, opened 2015-10-23
Fix code indent with 4 spaces for each levels

tagomoris:
There's some indentation rules for java code:
* Sun's code convensions: 2 spaces
 * http://www.oracle.com/technetwork/java/javase/documentation/codeconvtoc-136057.html
* Maven coding style: 4 spaces
 * https://maven.apache.org/developers/conventions/code.html
* Eclipse Coding Convention: tab (4 space wide)
 * https://wiki.eclipse.org/Coding_Conventions
* Google style guide: 2 spaces (4 spaces for continuation lines)
 * https://google.github.io/styleguide/javaguide.html#s4.5.2-line-wrapping-indent
* JRuby Style Guide (Java code): 4 spaces
 * https://github.com/jruby/jruby/wiki/JRubyStyleGuide

I think we should follow JRuby Style Guide.

PR #94, opened 2015-10-23
Fix to use modern ruby syntax to write code/tests

tagomoris:
@frsyuki Any thoughts?
I'll merge this 2 or 3 days later if there's no strong objections.
- https://github.com/xinminlabs/synvert
- https://github.com/xinminlabs/synvert-snippets

### DIFF
--- a/spec/cases_spec.rb
+++ b/spec/cases_spec.rb
@@ -1,7 +1,7 @@
 require 'spec_helper'
 require 'json'
 
-describe MessagePack do
+RSpec.describe MessagePack do
   here = File.dirname(__FILE__)
   CASES         = File.read("#{here}/cases.msg")
   CASES_JSON    = File.read("#{here}/cases.json")
@@ -16,7 +16,7 @@ describe MessagePack do
     js = JSON.load(CASES_JSON)
 
     ms.zip(js) {|m,j|
-      m.should == j
+      expect(m).to eq j
     }
   end
 
@@ -32,7 +32,7 @@ describe MessagePack do
     }
 
     ms.zip(cs) {|m,c|
-      m.should == c
+      expect(m).to eq c
     }
   end
 end
--- a/bench/pack.rb
+++ b/bench/pack.rb
@@ -4,7 +4,7 @@ require 'msgpack'
 data = { 'hello' => 'world', 'nested' => ['structure', {value: 42}] }
 data_sym = { hello: 'world', nested: ['structure', {value: 42}] }
 
-data = MessagePack.pack(:hello => 'world', :nested => ['structure', {:value => 42}])
+data = MessagePack.pack(hello: 'world', nested: ['structure', {value: 42}])
 
 Viiite.bench do |b|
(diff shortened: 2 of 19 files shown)


iconara:
Just as #93 this is so extremely unnecessary. Don't change code just to change code. It makes the git history meaningless. Please don't do this.

tagomoris:
`should` syntax will be disabled in future version of rspec, and msgpack-ruby now depends on `rspec ~> 3.3`.
https://github.com/rspec/rspec-expectations/blob/master/Should.md#using-either-expect-or-should-or-both

`should` syntax is not documented in `rspec-expectations` now, and it also doesn't work with objects which subclasses `BasicObject` (fortunately, we didn't use it for now).

tagomoris:
I'm not sure that I have enough time to fix problems at any day in future when Rspec change the default behavior and it makes all tests to fail.
So I want to change test code while I have some time to do it.

iconara:
There's no need to change as long as you don't update RSpec. Also, the old syntax is not going to go away. It's just not going to be enabled by default. That's a big difference.

tagomoris:
Rspec 3.3 dependency was introduced to use `skip` method to create pending specs.
Will NOT we use any new features in future? Will we reject all contributions with new rspec features or say "Rewrite it with old rspec feature or decline it"?

I want not to do such things.

iconara:
I'm saying update when it's necessary, don't do blanked rewrites of almost all of the test code just because it _might_ be needed some time in the future.

And I would absolutely reject contributions that rewrote all test code just to use a single new feature, especially when it's a tiny one like using `skip` instead of `pending`. I'm not even sure that there's a difference between those.

If you feel very strongly about this go ahead, I think it's unnecessary, but it's not nearly on the same level as #93.

spec/cases_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'spec_helper'
 require 'json'
 
-describe MessagePack do
+RSpec.describe MessagePack do
```
frsyuki: Is `Rspec.describe` recommended? I can't find about this syntax at http://betterspecs.org/#describe

frsyuki:
I added above comment but looks good for me overall. Apart from #93 which may be very different case, making code clean is generally good idea, in my opinion.
Then any contributors and committers can bring more with less effort. I don't say this is mandatory but it is good so that users can suppose msgpack-ruby is actively maintained OSS.
@tagomoris, I guessed that your point is that new code shouldn't to use old syntax which is considered obsoleted by rspec team, and new code is welcomed.
Ah, do you think it is not good idea, @iconara?

spec/cases_spec.rb
tagomoris: I'm not sure about it. It's introduced by `synvert` rule.
It looks to reduce `Kernel` modification dependency as much as possible. Not bad.

tagomoris: `rspec-core` README example uses `Rspec.describe` style.
https://github.com/rspec/rspec-core#basic-structure

iconara:
I'm closing this since it's not been merged in close to two years.

PR #96, opened 2015-10-23
update gh-pages document with the latest doclib/ code

frsyuki:
http://ruby.msgpack.org/ is generated from gh-pages branch. It was not updated...
Although it should be automated, I ran following commands for now:

```
bundle exec rake doc
git checkout gh-pages
cp doc/* ./ -a
```


tagomoris:
IMO, it's enough to add a section on README "How to update API References" with description @frsyuki wrote.

tagomoris:
Thanks :+1:

PR #97, opened 2015-10-24
README.md: added 'Updating documents' section

frsyuki:
Adding documents to update documents at http://ruby.msgpack.org. #96.


tagomoris:
:+1:

Issue #99, opened 2015-11-19
Serialized strings greater than 256 bytes fail with EOFError


ki4jnq:
Hey, Guys,

I've been looking to replace some existing Marshal code with MsgPack. But have been running into an issue when trying to serialize extended types.  If the serialized string is longer than 256 bytes, it will fail to unpack. I discovered the issue while working with some ActiveRecord, but I have reduced it to the following repeatable example.  I am running MRI v1.9.3-p551 on Ubuntu 14.04

```ruby
require 'msgpack'

class T
  def initialize(l)
    @l = l
  end

  def to_msgpack(packer=nil)
    packer ||= MessagePack::Packer.new
    packer.pack 'a'*@l
  end

  def self.from_msgpack(mp)
    MessagePack.unpack(mp.force_encoding("ASCII-8BIT"))
  end
end

MessagePack::DefaultFactory.register_type(0x01, T, packer: :to_msgpack, unpacker: :from_msgpack)

# This works
t1 = T.new(252)
s1 = MessagePack.pack t1
# s1 = "\xC7\xFE\x01\xD9\xFCaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
puts s1.length
# => 257
r1 =  MessagePack.unpack s1

# This fails
t2 = T.new(253)
s2 = MessagePack.pack t2
[...]

PR #100, opened 2015-11-20
fixes endian conversion of unpacking ext 16 and 32

frsyuki:
This fixes:

* unpacker doesn't convert endian of length part of ext 16 format
* unpacker doesn't convert endian of length part of ext 32 format
* packer uses ext 16 format when a length of body is 255 bytes

Fixes [#99].


tagomoris:
I added fix for JRuby implementation.

tagomoris:
I'm merging after CI passed.

PR #102, opened 2015-12-25
Unpacker can work with object non-IO but has #write

tagomoris:
In ruby way, Packer/Unpacker argument checks should be done only by `io.respond_to?("read")` and/or `io.respond_to?("write")`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ rvm:
   - 2.1.6
   - 2.2.2
   - ruby-head
-  - rbx-2
   - jruby-19mode
   - jruby-head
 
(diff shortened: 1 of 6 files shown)


tagomoris:
I added some stuffs about Rubinius:
* Rubinius runs super slowly with `cruby/*_spec.rb` tests, so I removed it from test target
* Tests on rubinius is failing in `bin/gem.rb` on TravisCI. I marked it as "allow_failure" temporally

Anyway, are there any users who want rubinius support? (I'll create another issue for that later)

frsyuki:
@tagomoris Did you mean `#read` in the tile of this pull-request?

frsyuki:
Looks good improvement because this improves API compatibility between CRuby and JRuby.

PR #103, opened 2016-01-06
Update rake-compiler-dock version to 0.5.0

cosmo0920:
`rake-compiler-dock` supports cross compiling against CRuby 2.3.0.
`rake-compiler-dock` pulls latest Docker image automatically.


cosmo0920:
Thanks!

PR #104, opened 2016-01-07
Improve Buffer compatibility about underlying io objects between cruby and jruby

tagomoris:
especially for the case used from Packer
* for Packer, Buffer requires #close, #flush and #write
* Buffer#read is not used by Packer/Unpacker, and read-only buffer should have only #read (w/o #write)

### DIFF
--- a/ext/java/org/msgpack/jruby/Buffer.java
+++ b/ext/java/org/msgpack/jruby/Buffer.java
@@ -42,8 +42,9 @@ public class Buffer extends RubyObject {
   @JRubyMethod(name = "initialize", optional = 2)
   public IRubyObject initialize(ThreadContext ctx, IRubyObject[] args) {
     if (args.length > 0) {
-      if (args[0].respondsTo("read") && args[0].respondsTo("write")) {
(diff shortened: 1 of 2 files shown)


tagomoris:
@iconara Could you review this change?

iconara:
Looks good to me. I would perhaps extract the now very long check for whether the object is IO-like to a method so that the code in `#initialize` was something like `if (isIoCompatible(io))` or `if (isIoLike(io))`. It's not a big deal though.

Another observation is that if the object fails the check it's silently ignored. That's not something that's new, and it's probably my fault from the beginning, but maybe we should do something about that.

tagomoris:
@iconara Your points look nice to me too. I think we may have to add `isIoLike` or something else in both of JRuby and CRuby, and also to raise exception for unexpected argument.
But it's a kind of compatibility change, and should be done with bumping minor version up because it breaks current working code with silently ignored non-io-like io arguments.l

tagomoris:
Merged.
About @iconara 's points, I filed #105.

PR #109, opened 2016-03-16
travis: Add jruby-9.0.5.0

cosmo0920:
JRuby encoding bug has been fixed.
It seems that this bug seems to be included JRuby 9.0.4.0 or earlier
major 9 versions.
jruby-9.0.0.0 and jruby-9.0.3.0 are also affected.


tagomoris:
Good catch. Thanks.

PR #111, opened 2016-04-19
Json2msgpack

nurse:
I write a conversion method from JSON to msgpack without creating internal Ruby objects.
Maybe it should release GVL to allow Ruby process another threads.


chrisseaton:
Shall we close this issue due to the extreme age?

What's the use-case?

It'd need rebasing otherwise, and then a Java implementation.

PR #115, opened 2016-05-25
fix SEGV that will happen with MRI 2.4.0 because of unification of Fixnum and Bignum into Integer

frsyuki:
MRI 2.4 will unify rb_cFixnum and rb_cBignum into rb_cInteger (rb_cFixnum == rb_cBignum). But underlaying implementation still require different function calls (FIXNUM_P / RBIGNUM_POSITIVE_P) depending on size of the integer. Thus msgpack-ruby needs to check difference using FIXNUM_P on runtime.
https://github.com/ruby/ruby/commit/f9727c12cc8fbc5f752f5983be1f14bb976e5a13

### DIFF
--- a/ext/msgpack/core_ext.c
+++ b/ext/msgpack/core_ext.c
@@ -65,6 +65,17 @@ static VALUE FalseClass_to_msgpack(int argc, VALUE* argv, VALUE self)
     return packer;
 }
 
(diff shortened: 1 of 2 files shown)


tagomoris:
LGTM

frsyuki:
Merged.

PR #116, opened 2016-05-25
Fixed RDoc about to_msgpack that actually requires packer

frsyuki:
IO is still supported for backward compatibility but they actually use
packer. This is because Packer has internal buffer and using IO as the
interface will require Packer to flush the buffer to IO every time.

### DIFF
--- a/doclib/msgpack/core_ext.rb
+++ b/doclib/msgpack/core_ext.rb
@@ -1,101 +1,101 @@
 
 class NilClass
   #
-  # Same as MessagePack.to_msgpack(self[, io]).
+  # Same as MessagePack.to_msgpack(self[, packer]).
   #
   # @return [String] serialized data
   #
-  def to_msgpack(io=nil)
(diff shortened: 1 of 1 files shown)


tagomoris:
`to_msgpack(packer) #=> packer` looks inconsistent API for me, hard to use/understand for users.
It is required to serialize objects without overhead of creating Strings for each objects serialized in recursive serialization (e.g, `Hash#to_msgpack`, `Array#to_msgpack`).

But, typically, `Object#to_xxx` returns instance of `xxx` in ruby. Returning packer is very confusing.

tagomoris:
I think that it's better to have another method to serialize objects recursively with a packer as an argument. It calls same method of child objects, and returns same packer object (or nil).
With that method, we can deprecate `to_msgpack` with arguments (the only valid API is `to_msgpack` without any arguments), in documents at least.

The biggest problem is: What is the best name for that method?

PR #114, opened 2016-05-13
Serializing symbols using an ext type

christopheraue:
re #110 

### How it's implementend

1. Packers and Factorys got a flag `has_symbol_ext_type`.
2. `has_symbol_ext_type` is set to `true` once an ext type for class `Symbol` is registered in their ext registry.
3. When encountering a symbol during serialization, the `has_symbol_ext_type` flag is checked. If it's `false` the symbol is serialized as string. If it's `true` its serialized as ext type.
4. `Symbol` got a default `#to_msgpack_ext` and `.from_msgpack_ext` so it does not need to be defined manually.

### Performance

This below benchmark compares the *master* branch with the *[preserve_symbols](https://github.com/christopheraue/msgpack-ruby/tree/preserve_symbols)* branch. With no ext type registered both have identical run times. Once the symbol ext type is registered, the custom serialization kicks in on the *preserve_symbols* branch and drags down performance:

Columns:

* bench:
  * `large_run`: 1 pack of an array of *count* symbols
  * `multi_run`: *count* packs of 1 symbol
* reg_type:
  * `none`: packers has no ext type registered
  * `symbol`: packer has an ext type registered for symbols

Benchmark:

    +------------+-----------+----------+------------------+-----------+----------+-----------+-----------+
    | :bench     | :reg_type | :count   | :branch          | :user     | :system  | :total    | :real     |
    +------------+-----------+----------+------------------+-----------+----------+-----------+-----------+
    | :large_run | :none     |   100000 | master           |  0.000000 | 0.000000 |  0.000000 |  0.002413 |
    | :large_run | :none     |   100000 | preserve_symbols |  0.000000 | 0.000000 |  0.000000 |  0.002356 |
    | :large_run | :none     |  1000000 | master           |  0.030000 | 0.000000 |  0.030000 |  0.031178 |
    | :large_run | :none     |  1000000 | preserve_symbols |  0.030000 | 0.000000 |  0.030000 |  0.029986 |
    | :large_run | :none     | 10000000 | master           |  0.820000 | 0.000000 |  0.820000 |  0.820698 |
    | :large_run | :none     | 10000000 | preserve_symbols |  0.810000 | 0.010000 |  0.820000 |  0.805589 |
    | :large_run | :symbol   |   100000 | master           |  0.000000 | 0.000000 |  0.000000 |  0.004980 |
    | :large_run | :symbol   |   100000 | preserve_symbols |  0.070000 | 0.000000 |  0.070000 |  0.070618 |
[...]

PR #122, opened 2016-06-27
Serializing symbols using an ext type

tagomoris:
This change follows #114, to add JRuby implementation.

I added two more fixes which is required to pass specs.
* Wrong spec to use ObjSpace to serialize/deserialize symbols: it's unsafe and unportable
* Bugfix to initialize ExtensionRegistry before setting it to Encoder: it's found by spec newly added by #114


tagomoris:
I'll merge this change after confirming CI green.

PR #124, opened 2016-07-18
Port Symbol extensions to C

metcalf:
7ea1cae7e3022ab04ee9f57ffb58ee12f1a6700b introduced an extension to Symbol to support easily using it as an ExtensionType. That implementation was in Ruby and incurred a good bit of performance overhead compared to native C code. This reimplements the same code in C.

### DIFF
--- a/ext/java/org/msgpack/jruby/MessagePackLibrary.java
+++ b/ext/java/org/msgpack/jruby/MessagePackLibrary.java
@@ -8,6 +8,8 @@ import org.jruby.RubyString;
 import org.jruby.RubyNil;
 import org.jruby.RubyBoolean;
 import org.jruby.RubyHash;
+import org.jruby.RubySymbol;
+import org.jruby.javasupport.JavaEmbedUtils;
 import org.jruby.runtime.load.Library;
 import org.jruby.runtime.builtin.IRubyObject;
 import org.jruby.runtime.ThreadContext;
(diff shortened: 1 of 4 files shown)


metcalf:
Ah, looks like I have some work to do for JRuby.

metcalf:
OK, all set on JRuby though since it's my first time writing anything for JRuby I'm sure there's a more elegant thing I'm missing.

tagomoris:
That methods are used when Symbol is registered to MessagePack::Factory without any serializer specifications (like `factory.register_type(0x30, Symbol)`).
IMO there are not so many use cases like that, and its implementation written in Ruby is good to show how symbols are serialized in easy way.

How do you think about it?

metcalf:
I hadn't considered the documentation advantages of the pure-Ruby implementation. I leave it up to your judgement as to whether you prefer the performance optimization or clarity of pure Ruby.

Deserializing large numbers of symbols is very expensive even when moved to C. `Marshal` maintains a symbol table when it serializes to reduce the cost of deserialization. I considered writing a subclass of Packer/Unpacker that maintains such a symbol table. Would you accept a PR to that effect if I get around to it?

tagomoris:
I'm not sure about how it improves performance, and also about how it makes code maintenance hard. So I cannot say anything right now. It depends on actual performance improvement and code complexity.

metcalf:
That's fair, I'll close this for now and let you re-open if you decide you want it. I suspect I won't actually get around to the symbol table code unless this becomes a major performance impediment again.

PR #125, opened 2016-08-01
Allow JRuby Unpacker to accept both a stream and options

leonid-shevtsov:
I found an inconsistency between the MRI and JRuby implementations:

```ruby
MessagePack::Unpacker.new(stream, symbolize_keys: true)
```

This code runs as expected on MRI/C implementation, but on JRuby it fails with an Unexpected number of arguments (2 for 1).

I've added a test case, and a fix.

(relevant failed test run on our gem https://circleci.com/gh/brightbytes/msgpack_rpc_client/11)


tagomoris:
LGTM!

PR #126, opened 2016-10-13
Fixed broken inheritance of ext types

christopheraue:
Currently, instead of getting back the packer proc of the superclass the superclass itself is returned on ext type lookup of subclasses. This leads to an `undefined method 'call' for <Class>` error.

The first commit gives variables in msgpack_packer_ext_registry_lookup a better name and makes the error obvious. The second commit fixes it.


tagomoris:
LGTM. Thank you for correcting that bug!

tagomoris:
I've just released v1.0.1 including this change. Thank you for contribution.

Issue #131, opened 2017-02-21
Extension type for modules


christopheraue:
Currently, an extension type can only be registered for classes: https://github.com/msgpack/msgpack-ruby/blob/master/ext/msgpack/packer_class.c#L283.

Couldn't modules be allowed as well? Inheritance of ext types works and already checks modules in the ancestor chain: https://github.com/msgpack/msgpack-ruby/blob/master/ext/msgpack/packer_ext_registry.h#L55

I'm looking to accomplish something like this (without inheritance):

```ruby
module Local
  def local_id
    # ...
  end
end

class LocalObject
  include Local
end

pk = MessagePack::Packer.new(io)
pk.register_type(0x01, Local) { |obj| obj.local_id }

obj1 = LocalObject.new
pk.write(obj1)

obj2 = LocalObject.new
pk.write(obj2)
```

And deserializing on a remote machine:

```ruby
class RemoteObject
  def initialize(remote_id)
    # ...
  end
end

uk = MessagePack::Unpacker.new(io)
uk.register_type(0x01) { |remote_id| RemoteObject.new(remote_id) }
```

tagomoris:
It looks reasonable for me.
@christopheraue can you write patches and tests for that update?

christopheraue:
Yes, I'll do that.

tagomoris:
`v1.1.0` has been released with this feature for all platform (including x86/64 & java).

PR #132, opened 2017-02-27
Enabled extension types for modules

christopheraue:
re #131 

What I've done is basically checking for class **or** module when registering an ext type for a packer of factory. Most of the noise comes from renaming variables accordingly and using `RubyModule` instead of `RubyClass` in jruby.

Finally, commit 0863031 adds support for serializing objects extended by a module with an ext type like `object.extend Mod`. To do this looking up an ext type in the ancestor chain already begins with the singleton class instead of its regular class.

### DIFF
--- a/ext/msgpack/packer.c
+++ b/ext/msgpack/packer.c
@@ -124,8 +124,18 @@ void msgpack_packer_write_hash_value(msgpack_packer_t* pk, VALUE v)
 void msgpack_packer_write_other_value(msgpack_packer_t* pk, VALUE v)
 {
     int ext_type;
-    VALUE proc = msgpack_packer_ext_registry_lookup(&pk->ext_registry,
-            rb_obj_class(v), &ext_type);
+
+    VALUE lookup_class;
+
+    if (SYMBOL_P(v)) { /* Symbols have no singleton class */
+      lookup_class = rb_obj_class(v);
+    } else {
+      lookup_class = rb_singleton_class(v);
+    }
+
+    VALUE proc = msgpack_packer_ext_registry_lookup(&pk->ext_registry, lookup_class,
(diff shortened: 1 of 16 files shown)


ext/msgpack/packer.c
```diff
@@ -124,8 +124,18 @@ void msgpack_packer_write_hash_value(msgpack_packer_t* pk, VALUE v)
 void msgpack_packer_write_other_value(msgpack_packer_t* pk, VALUE v)
 {
     int ext_type;
-    VALUE proc = msgpack_packer_ext_registry_lookup(&pk->ext_registry,
-            rb_obj_class(v), &ext_type);
+
+    VALUE lookup_class;
+
+    if (SYMBOL_P(v)) { /* Symbols have no singleton class */
```
tagomoris: This code checks only for Symbol objects, but there are some more classes to raise TypeError for `#singleton_class` method (e.g., Integer, Float).
Is it intended not to check these classes? I know that these classes should be packed in other functions, but i'm not sure there are no other classes which raises TypeError for `#singleton_class`.

christopheraue: I took a look at [`singleton_class_of`](https://github.com/ruby/ruby/blob/54a77d2044148bb442340f93747e1f345b6ca401/class.c#L1587) and the cases for which a TypeError is raised are Integer (Fixnum, Bignum), Float, Symbol and frozen String. Since all but symbols are already filtered out when reaching this code I left out the checks for them.

tagomoris: OK, great.
Could you add comments here to explain it, and to show where to see for the list of classes w/ TypeError?

tagomoris:
Merged. Thank you for great contribution!

PR #134, opened 2017-03-10
Register address so that DefaultFactory doesn't get GC'd

### LINKED ISSUE #133: Setting `DefaultFactory` to nil will cause a segv
Repro script:

```
$ ruby -Ilib -rmsgpack -e'MessagePack::DefaultFactory = nil; GC.start; "".to_msgpack'
```

This will cause a SEGV because [this global](https://github.com/msgpack/msgpack-ruby/blob/master/ext/msgpack/factory_class.c#L218) will get GC'd and the pointer will become bad.

tenderlove:
This program will SEGV without registering the global:

```
$ ruby -Ilib -rmsgpack -e'MessagePack::DefaultFactory = nil; GC.start; "".to_msgpack'
```

This fixes #133 by registering the address so that the object does not get GC'd


tagomoris:
I've merged #135 instead of this change.

PR #137, opened 2017-04-07
Add check for object's class on lookup. Fix #136

### LINKED ISSUE #136: incorrect ext type maybe used if an ancestor is also registered

ramonmaruko:


ext/msgpack/packer_ext_registry.h
```diff
@@ -59,8 +59,7 @@ static int msgpack_packer_ext_find_superclass(VALUE key, VALUE value, VALUE arg)
     return ST_CONTINUE;
 }
 
-
-static inline VALUE msgpack_packer_ext_registry_lookup(msgpack_packer_ext_registry_t* pkrg,
+static inline VALUE msgpack_packer_ext_registry_lookup_(msgpack_packer_ext_registry_t* pkrg,
```
tagomoris: Can you modify this name to more meaningful name?
IMO it's misleading.

tagomoris:
This change was merged into master via #152. Thank you!

PR #139, opened 2017-08-02
Bail out after a certain depth when packing.

simon-shopify:
This PR tackles the problem of failing fast on cyclical data when packing. On the current version of msgpack, packing in implemented with recursive C functions. This means that a value such as

```
cyclical = []
cyclical << cyclical
```

will trigger a stack overflow. If msgpack-ruby could detect this kind of data, it could bail out much faster. As far as know, I might also be safer since it won't rely on a memory protection error on a guard page while executing unmanaged code that we don't know whether it is safe to interrupt?

I went with the approach of having a `max_depth` option, which set the maximum depth of recursion. The alternative would be to keep a stack of visited `object_id` and raise if we see the same object twice.

Something else that could be done is to turn the recursive functions into iterative functions with a fixed-size stack on the size like it is done for decoding.

### DIFF
--- a/ext/msgpack/packer.c
+++ b/ext/msgpack/packer.c
@@ -27,6 +27,25 @@ static ID s_value;
 
 static ID s_call;
 
+static void msgpack_packer_increment_depth(msgpack_packer_t* pk)
+{
+    if(!pk->max_depth) {
+        return;
+    }
+
+    if(pk->depth >= pk->max_depth) {
+        rb_raise(rb_eArgError, "stack level too deep");
+    }
+
+    pk->depth += 1;
+}
+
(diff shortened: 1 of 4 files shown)


tagomoris:
IMO, self-recursive nested objects are result of bugs, and it's not ideal to get additional computing cost to detect such data structure in msgpack.
Otherwise, if your code produces self-recursive nested objects without any bugs, you should have checker of self-recursive structure out of msgpack.

But, possibly, I might miss other reasonable cases which requires such feature.
Any other opinion about this change?

simon-shopify:
I thought about it some more and I agree. I think I was being paranoid because of the stack overflow in unmanaged code.

Thanks for taking the time to look at it.

Issue #141, opened 2017-10-27
Feature Request: Fool-proof MessagePack::Factory#pack and #unpack methods


mwpastore:
I'd like to create a Factory instance to register a custom type and avoid polluting the DefaultFactory, but it's unclear how to use Factory#packer and Factory#unpacker in the simple case. It's definitely not as straightforward as MessagePack.pack and MessagePack.unpack. I see Packer#full_pack and Unpacker#full_unpack in the master branch, but they don't seem to be available in the current release. Here's what I've come up with:

```ruby
class Coder
  attr_reader :factory

  def initialize
    @factory = MessagePack::Factory.new
    @factory.register_type(0x00, Symbol)
  end

  def encode(data)
    factory.packer.write(data).to_str
  end

  def decode(data)
    factory.unpacker.feed(data).read
  end
end
```

For dumb-dumbs like me, it would be nice to have fool-proof Factory#pack and Factory#unpack (or Packer#dwim and Unpacker#dwim) methods that "do what I mean" for the simple cases. 

Failing that, maybe some nicer symmetry between the pack and unpack operations (why not call it e.g. `from_str` instead of `feed`)? 

Failing that, maybe some easier to follow documentation? 

Heck, even if it were just an option to pass your factory to MessagePack.pack and MessagePack.unpack, that would be great.

PR #142, opened 2017-12-01
Implement Factory#load and #dump for a more convenient API

casperisfine:
Fix: https://github.com/msgpack/msgpack-ruby/issues/141

I had the same issue than @mwpastore. IMO this simpler API should be enough for the vast majority of use cases:

```ruby
MY_CUSTOMER_PACKER = MessagePack::Factory.new
MY_CUSTOMER_PACKER.register_type(0x00, Symbol)

MY_CUSTOMER_PACKER.load(MY_CUSTOMER_PACKER.dump(:some_symbol))
```

### DIFF
--- a/lib/msgpack/factory.rb
+++ b/lib/msgpack/factory.rb
@@ -56,5 +56,26 @@ module MessagePack
         raise ArgumentError, "class or type id"
       end
     end
+
+    def load(src, param = nil)
+      unpacker = nil
+
+      if src.is_a? String
+        unpacker = unpacker(param)
+        unpacker.feed(src)
+      else
+        unpacker = unpacker(src, param)
+      end
+
+      unpacker.full_unpack
+    end
+    alias :unpack :load
+
+    def pack(v, *rest)
+      packer = packer(*rest)
+      packer.write(v)
+      packer.full_pack
+    end
+    alias :dump :pack
   end
 end
--- a/spec/factory_spec.rb
+++ b/spec/factory_spec.rb
@@ -44,6 +44,23 @@ describe MessagePack::Factory do
     end
   end
 
+  describe '#dump and #load' do
+    it 'can be used like a standard coder' do
+      subject.register_type(0x00, Symbol)
+      expect(subject.load(subject.dump(:symbol))).to be == :symbol
+    end
+
+    it 'is alias as pack and unpack' do
+      subject.register_type(0x00, Symbol)
+      expect(subject.unpack(subject.pack(:symbol))).to be == :symbol
+    end
+
+    it 'accept options' do
+      hash = subject.unpack(MessagePack.pack('k' => 'v'), symbolize_keys: true)
+      expect(hash).to be == { k: 'v' }
+    end
+  end
+
   class MyType
     def initialize(a, b)
       @a = a


tagomoris:
At first, we can't merge this request, because `Packer#pack` and `Unpacker#unpack` are already defined as aliases of `Packer#write` and `Unpacker#read`. This change breaks compatibility about these definitions.

I'm neutral about the idea to add `Packer#dump` and `Unpacker#load`.
@frsyuki How are you?

casperisfine:
> we can't merge this request, because Packer#pack and Unpacker#unpack are already defined as aliases of Packer#write and Unpacker#read

I'm defining `Factory#pack` and `Factory#unpack`, not `Packer / Unpacker`. But even though, I really only care about `Factory#load` and `Factory#dump`.

There is many APIs out there expected `load / dump`, so having `Factory` instances behave like a standard serializer is very valuable, e.g. with Rails:

```ruby
class MyModel < ActiveRecord::Base
  SERIALIZER = MessagePack::Factory.new
  SERIALIZER.register_type(0x00, Symbol)
  serialize :some_attribute, SERIALIZER
end
```

tagomoris:
Ah, I missread about that. Sorry. > `Factory#pack` & `Factory#unpack`
I agree that standard-like `#dump` and `#load` look valuable. IMO, we have two options to have such methods (if we're going to do it):
* add `#dump` and `#load` to Factory (generally speaking, factory class SHOULD NOT have such methods...)
* add a new proxy class to own `#dump` and `#load` (and `#pack`, `#unpack`?) (like `MessagePack`), and add a method to generate proxy in `Factory`

My second idea is like this:
```ruby
class MyModel < ActiveRecord::Base
  factory.register_type(0x00, Symbol)
  serialize :some_attribute, factory.proxy
end
```

casperisfine:
Sure I can do that. Not sure about `proxy` though. How about `serializer` ?

tagomoris:
I feel it's not bad > `Serializer` & `Factory#serializer`.
But, please wait. I want to hear from @frsyuki about that idea.

casperisfine:
Sure.

frsyuki:
This idea looks good.
@tagomoris Do you think that Factory class shouldn't have `#load` / `#dump` methods with a consideration that `MessagePack.load` and `MessagePack.dump` methods are the most frequently used methods? For me, it seems ok to add those methods to Factory class directly. We can think that they're alias of `.unpacker.load` or `.packer.dump` prepared for convenience.

tagomoris:
I think `Factory` should be just a factory for packer/unpacker in ideal situation - but I don't have an objection about adding `Factory#dump` and `#load` for convenience.
OK, I'll merge this request later - code looks good to me.

@casperisfine Just one more request: Could you create another pull-req to update API documents about these methods? https://github.com/msgpack/msgpack-ruby#updating-documents

casperisfine:
> Could you create another pull-req to update API documents about these methods?

Sure thing.

mwpastore:
The other idea I floated in #141 was the option of passing in your factory to MessagePack.pack and MessagePack.unpack, e.g.:

```ruby
packed_data = MessagePack.pack(data, :factory=>my_factory)
```

casperisfine:
@mwpastore the disadvantage of your solution is that it doesn't conform to the standard `dump(object)` and `load(string)` convention. So it requires glue code to be used in those cases.

tagomoris:
I agree about the disadvantage mentioned by @casperisfine of @mwpastore 's idea. I think that the code in this request provides max convenience for users with minimum code.

tagomoris:
Merged! I'll release a new version gem after I'll add an another patch about `Factory`.

tagomoris:
`v1.2.0` has been released with this change. Thank you!

PR #143, opened 2017-12-05
Generate doc for Factory#load and Factory#dump

casperisfine:
Doc update for https://github.com/msgpack/msgpack-ruby/pull/142 as requested.

cc @tagomoris

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ Gemfile*
 *~
 /msgpack-ruby
 /vendor
+/.yardoc
+/doc
--- a/Array.html
+++ b/Array.html
@@ -117,7 +117,7 @@
         <li class="public ">
   <span class="summary_signature">
     
-      <a href="#to_msgpack-instance_method" title="#to_msgpack (instance method)">- (String) <strong>to_msgpack</strong>(io = nil) </a>
+      <a href="#to_msgpack-instance_method" title="#to_msgpack (instance method)">- (String) <strong>to_msgpack</strong>(packer = nil) </a>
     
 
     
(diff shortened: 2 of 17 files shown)


tagomoris:
@casperisfine Ah, I'm sorry that I missed what I should tell you...
The API documents of this library are written under `doclib` in `master` (and other development) branch. So the exact list of what we need to do when we want to update API document to add new methods is (for example, add new methods to `MessagePack::Factory`):

1. update `doclib/msgpack/factory.rb` to add methods
2. create a pull-req to master branch with the diff above
3. execute `bundle exec rake doc` with the diff above to generate updated documents under `doc`
4. make a pull-req to `gh-pages` with contents under `doc` (by copying it to root)

casperisfine:
I'll get right on that!

.gitignore
```diff
@@ -5,3 +5,5 @@ Gemfile*
 *~
 /msgpack-ruby
 /vendor
+/.yardoc
+/doc
```
casperisfine: I took the liberty to add those 2, it makes the process much simpler.

casperisfine:
Weird. I ran `rake doc` from https://github.com/msgpack/msgpack-ruby/pull/145, but it doesn't seem to generate doc for the methods I added 🤔

tagomoris:
I'm supposing that it's caused by this blank line: https://github.com/msgpack/msgpack-ruby/pull/145#discussion_r155202275

casperisfine:
I fixed it, but still no luck :/

tagomoris:
I created #146 and merged it because it seems to work well.
@casperisfine  Thank you!

PR #144, opened 2017-12-06
Freeze MessagePack::Factory correctly

tagomoris:
Now we can register new types to frozen `MessagePack::Factory`.
It's clearly a bug, because we may get different results from `Factory#dump` or `Factory#load` after `Factory#freeze`.
We should reject type registration operation for frozen Factory objects.


tagomoris:
@frsyuki PTAL

frsyuki:
@tagomoris Looks perfect!
It's a bit surprising that Ruby doesn't have a specific error class for modifying frozen objects

PR #145, opened 2017-12-06
Document Factory#dump and Factory#load

casperisfine:
I mostly compiled the relevant parts from the existing methods docs.

cc @tagomoris

### DIFF
--- a/doclib/msgpack/factory.rb
+++ b/doclib/msgpack/factory.rb
@@ -20,6 +20,22 @@ module MessagePack
     end
 
     #
+    # Serialize the passed value
+    #
+    # If it could not serialize the object, it raises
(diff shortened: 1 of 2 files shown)


doclib/msgpack/factory.rb
```diff
@@ -19,6 +19,22 @@ def initialize
...
+    #
+    # Serialize the passed value
+    #
+    # If it could not serialize the object, it raises
+    # NoMethodError: undefined method `to_msgpack' for #<the_object>.
+    #
+    # @param obj [Object] object to serialize
+    # @param options [Hash]
+    # @return [String] serialized object
+
```
tagomoris: Weird blank line - missing `#`?

casperisfine: Fixed!

PR #146, opened 2017-12-07
API document update

tagomoris:
This branch is to update API documents from current master branch, which should have `MessagePack::Factory#load` and `MessagePack::Factory#dump`.

See #143.


tagomoris:
I followed the steps to regenerate API documents from current master, and got expected results.
I couldn't know why #143 doesn't work well 🤔  ... anyway, I'll merge this into `gh-pages`.

PR #147, opened 2017-12-07
Simplify the list of ruby to execute CI

tagomoris:
CI builds of current master branch are stuck by osx builds, and there seems no needs to run so many osx builds.
So I'm removing osx builds for older versions of Ruby, and update configuration file to specify the list to be executed (instead of `versions * os - exclude_list`) to simplify it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,5 @@
 language: ruby
 
-rvm:
-  - 2.0.0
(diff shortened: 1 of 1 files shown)


tagomoris:
The current status of CI build for master branch:

tagomoris:
After this change:

Issue #150, opened 2018-01-03
sometimes segfaulting with ruby 2.5.0


doits:
Sometime after upgrade to ruby 2.5.0 I hit the following segfault:

```
[BUG] Segmentation fault at 0x000000000000003f
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0019 p:---- s:0111 e:000110 CFUNC  :packer
c:0018 p:0014 s:0107 e:000106 METHOD /cache/gems/ruby_2.5/gems/msgpack-1.2.0/lib/msgpack.rb:42
c:0017 p:0148 s:0100 e:000097 METHOD /cache/gems/ruby_2.5/gems/bootsnap-1.1.7/lib/bootsnap/load_path_cache/store.rb:75
c:0016 p:0009 s:0092 e:000091 METHOD /cache/gems/ruby_2.5/gems/bootsnap-1.1.7/lib/bootsnap/load_path_cache/store.rb:55
c:0015 p:0028 s:0088 e:000086 METHOD /cache/gems/ruby_2.5/gems/bootsnap-1.1.7/lib/bootsnap/load_path_cache/store.rb:47
c:0014 p:0007 s:0083 e:000082 METHOD /cache/gems/ruby_2.5/gems/bootsnap-1.1.7/lib/bootsnap/load_path_cache/cache.rb:138
c:0013 p:0057 s:0078 e:000077 BLOCK  /cache/gems/ruby_2.5/gems/bootsnap-1.1.7/lib/bootsnap/load_path_cache/cache.rb:121 [FINISH]
c:0012 p:---- s:0075 e:000074 CFUNC  :synchronize
c:0011 p:0012 s:0071 e:000070 METHOD /cache/gems/ruby_2.5/gems/bootsnap-1.1.7/lib/bootsnap/load_path_cache/cache.rb:115
c:0010 p:0064 s:0066 e:000065 METHOD /cache/gems/ruby_2.5/gems/bootsnap-1.1.7/lib/bootsnap/load_path_cache/cache.rb:14 [FINISH]
c:0009 p:---- s:0058 e:000057 CFUNC  :new
c:0008 p:0030 s:0051 e:000050 METHOD /cache/gems/ruby_2.5/gems/bootsnap-1.1.7/lib/bootsnap/load_path_cache.rb:29
[...]

PR #151, opened 2018-01-10
Fix SEGV on hash in ext-registry

tagomoris:
The current code allocates structs in `_alloc` methods and creates Hash VALUEs at the same method.
But these VALUEs are not marked until `#initialize` runs, and if GC runs before `#initialize`, these VALUEs will be collected. Generally, VALUEs should be created/assigned at `#initialize`.

On the other hand, other SEGV occurs if we simply move code to create VALUEs to `#initialize`. This is caused by GC, which refers the allocated/uninitialized memory area of structs (allocated by `ALLOC_N`).
@ko1 [said](https://twitter.com/_ko1/status/950973448284549120) that "ALLOC_N brings death to you." We need to use `ZALLOC_N` everywhere.

This change is to fix these problems found (and possibly danger code), and to add a test to make SEGV without these patches.

### DIFF
--- a/ext/msgpack/buffer_class.c
+++ b/ext/msgpack/buffer_class.c
@@ -54,7 +54,7 @@ static void Buffer_free(void* data)
 
 static VALUE Buffer_alloc(VALUE klass)
 {
-    msgpack_buffer_t* b = ALLOC_N(msgpack_buffer_t, 1);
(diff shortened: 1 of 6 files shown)


tagomoris:
@frsyuki Could you review this change?

tagomoris:
I think that this should solve #150.

PR #152, opened 2018-01-10
Fix class lookup in packer registry

### LINKED ISSUE #136: incorrect ext type maybe used if an ancestor is also registered

tagomoris:
Fixes #136, follows #137.

#137 has difficulties to be merged, but it shows that packer registry lookup rule has a bug for super-and-sub classes.
This change includes the fix of #137, and add/fix some tests, comments and change of function names.


tagomoris:
@frsyuki PTAL

tagomoris:
Oops. I'll add commits for JRuby, and will merge this after CI green.

tagomoris:
Green!

PR #161, opened 2019-01-20
Add missing require and set Encoding.default_external

eregon:
* Otherwise, `bundle exec rspec spec/cruby/buffer_spec.rb` fails.


eregon:
This might be something that is best added in `spec_helper.rb`.
I followed the existing convention for a small change and checked all spec files can be executed individually.

eregon:
Thanks for merging!

tagomoris:
Thank you for these contributions! I'm looking forward to hearing what you did in details at conferences :D

PR #163, opened 2019-02-19
Add write_bin_header

ojundt:
Hi,

I was trying to append a large binary file to a msgpack string in streaming style without loading the binary file into memory and passing it to `Packer#write` but I couldn't find a way. I'm missing a `write_bin_header` method, similar to `write_array_header` and `write_map_header`.

Luckily there's already a `msgpack_packer_write_bin_header(length)` method. This made adding the ruby wrapper methods a piece of cake :)

### DIFF
--- a/doclib/msgpack/packer.rb
+++ b/doclib/msgpack/packer.rb
@@ -92,6 +92,12 @@ module MessagePack
     end
 
     #
+    # Serializes a string object as binary data. Same as write("string".encode(Encoding::BINARY)).
+    #
+    def write_bin(obj)
+    end
+
+    #
     # Write a header of an array whose size is _n_.
     # For example, write_array_header(1).write(true) is same as write([ true ]).
     #
@@ -116,7 +122,7 @@ module MessagePack
     # io.write('chunk1')
     # io.write('chunk2')
     # is the same as
-    # write('chunk1chunk2'.force_encoding(Encoding::BINARY)).
+    # write('chunk1chunk2'.encode(Encoding::BINARY)).
     #
     # @return [Packer] self
     #
--- a/doclib/msgpack/packer.rb
+++ b/doclib/msgpack/packer.rb
@@ -110,6 +110,20 @@ module MessagePack
     end
 
     #
+    # Write a header of a binary string whose size is _n_. Useful if you want to append large binary data without loading it into memory at once.
+    # For example,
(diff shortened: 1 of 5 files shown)


tagomoris:
To follow the existing methods manner, we should support `Packer#write_bin` in the same way with `Packer#write_string` instead of `Packer#write_bin_header`.
`#write_xxx_header` methods are supported only for container-like data structure, `map` and `array` ("container-like" means that structure contains other msgpack-type values). Other msgpack-types are supported by `write_xxx` directly, not only for headers.

The missing `Packer#write_bin` should be supported.
@ojundt Could you update to add a method `Packer#write_bin`? Or do you have any reason to prefer `#write_bin_header` rather than `#write_bin`?

ojundt:
`Packer#write_bin(value)` would still require `value` to be fully loaded into memory. That's what I'm trying to prevent. The `write_bin_header` method has the advantage that you are totally free in how you append the binary data to the output.

In my specific case I have a method that should yield chunks of msgpacked binary data to the given block. The block uses the chunks to write to a file or a HTTP stream or some other output stream.

With this PR my implementation looks like this:
```
def msgpack_each
  packer = MessagePack::Packer.new
  packer.write_map_header(2)
  packer.write(:some_metadata_header).write(@some_value)

  @input_file.open

  packer.write(:input_file).write_bin_header(@input_file.size)
  yield(packer.to_s)

  while (chunk = @input_file.read(8192))
    yield(chunk)
  end
ensure
  @input_file.close
end
```

tagomoris:
@ojundt just to confirm: your important motivation is to reduce times of copies of target binary data, especially in the case of large data, is it?

If so, it’s reasonable to wish #write_bin_header method. On the other hand, I’m still thinking that it’s better to have #write_bin too for API consistency. 
Can I ask you to add it?

ojundt:
Done!

doclib/msgpack/packer.rb
```diff
@@ -91,6 +91,12 @@ def write(obj)
     def write_nil
     end
 
+    #
+    # Serializes a string object as binary data. Same as write("string".encode(Encoding::BINARY)).
+    #
+    def write_nil
```
tagomoris: It's duplicated. Is it `def write_bin`?

ojundt: fixed

tagomoris:
Merged, and released `v1.2.7`. Thank you!

PR #168, opened 2019-06-17
MessagePack timestamp type

gfx:
This PR implements the timestamp type of MessagePack: timestamp32, imestamp64, and timestamp96. It also registers Time as the timestamp type for DefaultFactory.

## MessagePack::Timestamp class

This is the core logic for the timestamp type.

This is independent from the built-in Time class because  I'll  use the timestamp type in a Rails app,  overriding it by `MessagePack::DefaultFactory.register_type(MessagePack::TImestamp::TYPE, ActiveSupport::TimeWithZone)`.

## Default behavior

This PR does not change the default behavior because it is not a trivial thing.

However, `factory.register_type(-1, Time)` is tested.

cf. d31e669

## Specification

https://github.com/msgpack/msgpack/blob/master/spec.md

### DIFF
--- a/lib/msgpack.rb
+++ b/lib/msgpack.rb
@@ -23,8 +23,6 @@ module MessagePack
   DefaultFactory = MessagePack::Factory.new
   DEFAULT_EMPTY_PARAMS = {}.freeze
 
-  DefaultFactory.register_type(Timestamp::TYPE, Time)
-
   def load(src, param = nil)
     unpacker = nil
 
--- a/lib/msgpack.rb
+++ b/lib/msgpack.rb
@@ -17,6 +17,7 @@ require "msgpack/unpacker"
 require "msgpack/factory"
 require "msgpack/symbol"
 require "msgpack/core_ext"
+require "msgpack/timestamp"
 
 module MessagePack
   DefaultFactory = MessagePack::Factory.new
--- a/lib/msgpack.rb
+++ b/lib/msgpack.rb
@@ -23,6 +23,8 @@ module MessagePack
   DefaultFactory = MessagePack::Factory.new
   DEFAULT_EMPTY_PARAMS = {}.freeze
 
+  DefaultFactory.register_type(Timestamp::TYPE, Time)
+
   def load(src, param = nil)
     unpacker = nil
 
--- a/lib/msgpack.rb
+++ b/lib/msgpack.rb
@@ -18,6 +18,7 @@ require "msgpack/factory"
 require "msgpack/symbol"
 require "msgpack/core_ext"
 require "msgpack/timestamp"
+require "msgpack/time"
 
 module MessagePack
   DefaultFactory = MessagePack::Factory.new
--- a/lib/msgpack/core_ext.rb
+++ b/lib/msgpack/core_ext.rb
@@ -126,6 +126,19 @@ else
   end
 end
 
+class Time
+  include MessagePack::CoreExt
+
+  def self.from_msgpack_ext(payload)
+    tv = MessagePack::Timestamp.from_msgpack_ext(payload)
+    at(tv.sec, tv.nsec, :nanosecond)
+  end
+
+  def to_msgpack_ext
+    MessagePack::Timestamp.to_msgpack_ext(tv_sec, tv_nsec)
+  end
+end
+
 module MessagePack
   class ExtensionValue
     include CoreExt
--- a/lib/msgpack/core_ext.rb
+++ b/lib/msgpack/core_ext.rb
@@ -13,7 +13,7 @@ module MessagePack
     end
   end
 
-  # for namespacing
+  # 3-arg Time.at is available Ruby >= 2.5
   TIME_AT_3_AVAILABLE = begin
     Time.at(0, 0, :nanosecond)
     true
--- a/lib/msgpack/core_ext.rb
+++ b/lib/msgpack/core_ext.rb
@@ -12,6 +12,14 @@ module MessagePack
       end
     end
   end
+
+  # for namespacing
+  TIME_AT_3_AVAILABLE = begin
+    Time.at(0, 0, :nanosecond)
+    true
+  rescue ArgumentError
+    false
+  end
 end
 
 class NilClass
@@ -131,7 +139,12 @@ class Time
 
   def self.from_msgpack_ext(payload)
     tv = MessagePack::Timestamp.from_msgpack_ext(payload)
-    at(tv.sec, tv.nsec, :nanosecond)
+
+    if MessagePack::TIME_AT_3_AVAILABLE
+      at(tv.sec, tv.nsec, :nanosecond)
+    else
+      at(tv.sec, tv.nsec / 1000.0)
+    end
   end
 
   def to_msgpack_ext
--- a/lib/msgpack/core_ext.rb
+++ b/lib/msgpack/core_ext.rb
@@ -12,14 +12,6 @@ module MessagePack
       end
     end
   end
-
-  # 3-arg Time.at is available Ruby >= 2.5
-  TIME_AT_3_AVAILABLE = begin
-    Time.at(0, 0, :nanosecond)
-    true
-  rescue ArgumentError
-    false
-  end
 end
 
 class NilClass
@@ -131,24 +123,6 @@ else
       packer.write_int self
       packer
     end
-  end
-end
-
-class Time
-  include MessagePack::CoreExt
-
-  def self.from_msgpack_ext(payload)
-    tv = MessagePack::Timestamp.from_msgpack_ext(payload)
-
-    if MessagePack::TIME_AT_3_AVAILABLE
-      at(tv.sec, tv.nsec, :nanosecond)
-    else
-      at(tv.sec, tv.nsec / 1000.0)
-    end
-  end
-
-  def to_msgpack_ext
-    MessagePack::Timestamp.to_msgpack_ext(tv_sec, tv_nsec)
   end
 end
 
--- a/lib/msgpack/time.rb
+++ b/lib/msgpack/time.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# MessagePack extention packer and unpacker for built-in Time class
+module MessagePack
+  module Time
+    # 3-arg Time.at is available Ruby >= 2.5
+    TIME_AT_3_AVAILABLE = begin
+      !!::Time.at(0, 0, :nanosecond)
+    rescue ArgumentError
+      false
+    end
+
+    Unpacker = if TIME_AT_3_AVAILABLE
+                 lambda do |payload|
+                   tv = MessagePack::Timestamp.from_msgpack_ext(payload)
+                   ::Time.at(tv.sec, tv.nsec, :nanosecond)
+                 end
+               else
+                 lambda do |payload|
+                   tv = MessagePack::Timestamp.from_msgpack_ext(payload)
+                   ::Time.at(tv.sec, tv.nsec / 1000.0)
+                 end
+               end
+
+    Packer = lambda { |time|
+      MessagePack::Timestamp.to_msgpack_ext(time.tv_sec, time.tv_nsec)
+    }
+  end
+end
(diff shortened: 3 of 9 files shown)


lib/msgpack.rb
```diff
@@ -17,11 +17,14 @@
 require "msgpack/factory"
 require "msgpack/symbol"
 require "msgpack/core_ext"
+require "msgpack/timestamp"
 
 module MessagePack
   DefaultFactory = MessagePack::Factory.new
   DEFAULT_EMPTY_PARAMS = {}.freeze
 
+  DefaultFactory.register_type(Timestamp::TYPE, Time)
```
gfx: I'm not sure this is good, but at least it is useful for testing.

lib/msgpack/timestamp.rb
```diff
@@ -0,0 +1,54 @@
...
+    TIMESTAMP64_MAX_SEC = 0x400000000 # 34-bit unsigned int
+
+    def self.from_msgpack_ext(data)
+      case data.length
+      when 4
+        # timestamp32 = (uint32be)
+        sec = data.unpack('L>').first
+        new(sec, 0)
+      when 8
+        # timestamp64 = (uint32be, uint32be)
```
tagomoris: This comment is misleading... actually it's `(uint30be, uint34be)`.
I understood that `(uint32be, uint32be)` is the format to unpack the binary, but it confused me for a while.

lib/msgpack/timestamp.rb
```diff
@@ -0,0 +1,54 @@
...
+        n, s = data.unpack('L>2')
+        sec = (n & 0x3) * 0x100000000 + s
```
tagomoris: Can you use `0b11` as the mask and `<< 32` instead of `* 0x100000000`?
It's easy to read for me because my recommendation shows "lower 2bit mask" and "use it as upper bytes". Counting the number of zero in `0x100000000` is scary.

tagomoris: And... this is the just an idea, which might be a kind of too-early optimization.
How about this code? `sec = (n == 0 ? s : (n & 0b11 << 32) + s)`
Until 2038, `n` is 0 always (or in most cases).

lib/msgpack/timestamp.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module MessagePack
+  Timestamp = Struct.new(:sec, :nsec)
+
+  class Timestamp # a.k.a. "TimeSpec"
+    TYPE = -1 # timestamp extension type
+
+    TIMESTAMP32_MAX_SEC = 0x100000000 # 32-bit unsigned int
```
tagomoris: IMO "max" means the inclusive upper limit. For example, the max of `[0, 99]` is 99, but 98 for `[0, 99)`.
The max value of 32-bit bytes looks the 8 chars of `f`, not 9 chars.
How do you feel about it?

lib/msgpack/timestamp.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def self.to_msgpack_ext(sec, nsec)
+      if sec >= 0 && nsec >= 0 && sec < TIMESTAMP64_MAX_SEC
+        if nsec === 0 && sec < TIMESTAMP32_MAX_SEC
+          # timestamp32 = (uint32be)
+          [sec].pack('L>')
+        else
+          # timestamp64 (uint32be, uint32be)
+          sec_high = sec << 32
```
tagomoris: I have the same idea to deserialization: What do you think to have `if sec < TIMESTAMP32_MAX_SEC` here?

lib/msgpack/timestamp.rb
```diff
@@ -0,0 +1,54 @@
...
+    TYPE = -1 # timestamp extension type
+
+    TIMESTAMP32_MAX_SEC = 0x100000000 # 32-bit unsigned int
+    TIMESTAMP64_MAX_SEC = 0x400000000 # 34-bit unsigned int
+
+    def self.from_msgpack_ext(data)
+      case data.length
+      when 4
+        # timestamp32 = (uint32be)
+        sec = data.unpack('L>').first
```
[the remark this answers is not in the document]
tagomoris: Let's have it as comments!

lib/msgpack/core_ext.rb
```diff
@@ -126,6 +134,24 @@ def to_msgpack_with_packer(packer)
   end
 end
 
+class Time
+  include MessagePack::CoreExt
+
+  def self.from_msgpack_ext(payload)
```
tagomoris: This may break existing user code if users defined their own `Time.from_msgpack_ext` (and `Time#to_msgpack_ext`).
It seems better not to have these methods, and to add options to enable predefined types on Factory class. What do you think about this idea?
(It requires patches for C/Java ext, so I'm ok to do it by myself, out of this change)

lib/msgpack/timestamp.rb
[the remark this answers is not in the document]
gfx: Hmm, but it's just a built-in feature of Ruby. "uint32be" isn't enough?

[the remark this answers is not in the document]
tagomoris: Probably yes!
I don't have a strong opinion, but "If you want to add notes on a pull-req beforehand, it should be done as code comments".

lib/msgpack/timestamp.rb
gfx: I think its optimization is a bit more  complicated like this:

```ruby
sec_hi_2bits = n & 0b11
sec = sec_hi_2bits == 0 ? s : (sec_hi_2bits << 32) + s
```

I'm not sure it is faster than just `(n & 0b11) << 32) + s`.

lib/msgpack/timestamp.rb
gfx: Yep, it's confusing. `(1 << 32) - 1` is absolutely better.

lib/msgpack/timestamp.rb
tagomoris: Ah, your code looks right. And here's my microbenchmark:
```ruby
require 'benchmark'

t = Time.now
n = t.nsec << 2
s = t.sec

times = 1_000_000
Benchmark.bm do |x|
  x.report(:simple) do
    for i in 1..times
      sec = (n & 0b11 << 32) + s
    end
  end
  x.report(:optimized) do
    for i in 1..times
      sec_hi_2bits = n & 0b11
      sec = sec_hi_2bits == 0 ? s : (sec_hi_2bits << 32) + 2
    end
  end
end
```
The result was:
```
       user     system      total        real
simple  0.055794   0.000129   0.055923 (  0.056197)
optimized  0.047181   0.000082   0.047263 (  0.047393)
```
Let's forget this idea...

lib/msgpack/timestamp.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module MessagePack
+  Timestamp = Struct.new(:sec, :nsec)
+
+  class Timestamp # a.k.a. "TimeSpec"
```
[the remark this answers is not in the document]
gfx: I can't find how to use the `Struct.new BLOCK` syntax with constants:

NG(1):

```
$ ruby -e 'C = Struct.new(:foo) do CONST = 42 end; pp(C::CONST)'
Traceback (most recent call last):
-e:1:in `<main>': uninitialized constant C::CONST (NameError)
```

NG(2):

```
$ ruby -e 'Struct.new("::C", :foo) do pp self; CONST = 42 end; pp(C::CONST)'
Traceback (most recent call last):
	1: from -e:1:in `<main>'
-e:1:in `new': identifier ::C needs to be constant (NameError)
```

NG(3; it's not what I want):

```
$ ruby -e 'Struct.new("C", :foo) do pp self; self::CONST = 42 end; pp(Struct::C::CONST)'
Struct::C
42
```
I'm sorry I'm not good at Ruby 😭

lib/msgpack/timestamp.rb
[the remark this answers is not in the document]
gfx: Added comments about the big-endian specifier with a link to the Ruby reference!

lib/msgpack/time.rb
```diff
@@ -0,0 +1,29 @@
...
+
+# MessagePack extention packer and unpacker for built-in Time class
+module MessagePack
+  module Time
+    # 3-arg Time.at is available Ruby >= 2.5
+    TIME_AT_3_AVAILABLE = begin
+      !!::Time.at(0, 0, :nanosecond)
+                          rescue ArgumentError
+                            false
+    end
```
gfx: FYI the above crazy indentation is made by rubocop.

lib/msgpack/timestamp.rb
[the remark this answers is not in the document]
tagomoris: Do you mean that there is no way to refer `MessagePack::Timestamp::TYPE` from other classes?
It's not impossible (for example, `C = Struct.new(); C.const_set(:CONST, 42); p C::CONST`), but I guess we should define normal modules/classes instead of using Struct.
Is there any reason to use Struct?

lib/msgpack/time.rb
tagomoris: How about this code?
```ruby
TIME_AT_3_AVAILABLE = !!::Time.at(0, 0, :nanosecond) rescue false
```

lib/msgpack/timestamp.rb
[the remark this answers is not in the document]
gfx: I use Struct because of its auto-generated `#==` method, which is useful only for testing.

But yes, I'll use a normal class; it's just easy.

lib/msgpack/timestamp.rb
gfx: I'm not sure what you said. I think it's already checked the above `if` condition.

lib/msgpack/core_ext.rb
gfx: The latest code no longer break the existing code: Instead, use `MessagePack::Time::{Packer,Unpacker}` by hand:

```ruby
      factory.register_type(
        MessagePack::Timestamp::TYPE,
        Time,
        packer: MessagePack::Time::Packer,
        unpacker: MessagePack::Time::Unpacker
      )
```

as tests show.

lib/msgpack/timestamp.rb
tagomoris: Oh, sorry. This comment was about the optimization idea until 2038. Nevermind.

lib/msgpack/timestamp.rb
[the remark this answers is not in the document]
gfx: done: 733ed14

lib/msgpack/time.rb
gfx: rubocop fixes it to:

```
    TIME_AT_3_AVAILABLE = begin
                            !!::Time.at(0, 0, :nanosecond)
                          rescue StandardError
                            false
                          end
```

but yes, it's better than the current, so apply it f49a8ff

I wonder the rubocop behavior. It sometimes seems to change the code at random.

gfx:
Everything is done!

BTW should I add documents about it in this PR?

tagomoris:
@gfx It's better to have if possible!

I'm thinking about the option to move code in `MessagePack::Time` to `Timestamp` (those two look confusing to me...), but I'll add fixes on those at the same time with the switch to enable predefined ext types on Factory.
So, this change looks good to me. I'll merge this after the commit for docs.

gfx:
I think MessagePack::Time and MessagePack::Timestamp are completely different classes.

MessagePack::Timestamp provides utilities for implementing custom timestamp mapping.

MessagePack::Time provides the packer and unpacker for built-in Time class.  IIf msgpack-ruby supports ActiveSupport::TImeWithZone, its name will be MessagePack::ActiveSupportTimeWithZone or something alike.

gfx:
doc is added! 7566db5

tagomoris:
I've added some tests 6c89d3b0a534537931bedfc0b31634942250b5fa and released `v1.3.0`.
Thank you for the great patches!

PR #170, opened 2019-06-26
Fix misspelled word: Unpcker -> Unpacker

wcmonty:
I noticed that `Unpacker` was not spelled consistently while going through the diff of the last release.


tagomoris:
Good catch. Thanks!

gfx:
Sorry! 😭

PR #172, opened 2019-08-03
fix timestamp64 bit ops

gfx:
Fix a stupid mistake 😭 

reported by https://twitter.com/otake84/status/1157592556386996224


tagomoris:
Shipped as `v1.3.1`.

Issue #173, opened 2019-10-09
Fix (Potentially) Unnecessary T_HASH Check  


martelogan:
I could be mistaken but there seems to be an unnecessary double-check for `T_HASH` here:

https://github.com/msgpack/msgpack-ruby/blob/b48f9580bfd33be6a86652f10a41bd691a7d0c91/ext/msgpack/unpacker_class.c#L78-L80

Why is the second check required when we've already tested the type of `v` against `T_HASH`, then only rebound it to `options` ? 

I could be missing something here since I notice [`v0.5.8`](https://github.com/msgpack/msgpack-ruby/releases/tag/v0.5.8) added it when working to [add the `:symbolize_keys` option](https://github.com/msgpack/msgpack-ruby/commit/0854d61673239d39a572e47d9b0734103b9e4b5b) as mentioned by

https://github.com/msgpack/msgpack-ruby/blob/acc07af4b3a8a4b0384c8e84e549d580fefd02b8/ChangeLog#L151-L154

PR #174, opened 2019-10-09
Fix unnecessary T_HASH check

martelogan:
### Goal

Resolves https://github.com/msgpack/msgpack-ruby/issues/173

I could be mistaken but there seems to be an unnecessary double-check for `T_HASH` here:

https://github.com/msgpack/msgpack-ruby/blob/b48f9580bfd33be6a86652f10a41bd691a7d0c91/ext/msgpack/unpacker_class.c#L78-L80

Why is the second check required when we've already tested the type of `v` against `T_HASH`, then only rebound it to `options` ? 

### Approach

Simply remove the (seemingly) unnecessary second-check: https://github.com/msgpack/msgpack-ruby/commit/0b1bf902375283985fb0963143e5f9528b2767df

### Attn

Feel free to merge if I'm correct, otherwise: I'm just interested to learn why this check was required :p

### DIFF
--- a/ext/msgpack/unpacker_class.c
+++ b/ext/msgpack/unpacker_class.c
@@ -77,9 +77,6 @@ VALUE MessagePack_Unpacker_initialize(int argc, VALUE* argv, VALUE self)
         VALUE v = argv[0];
         if(rb_type(v) == T_HASH) {
             options = v;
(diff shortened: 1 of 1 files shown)


tagomoris:
That's should be a very simple mistake. Thank you for reporting!

PR #185, opened 2020-04-22
Address Rubocop Issues

benhayehudi:
Rubocop is a fast-moving tool that incorporates many breaking changes in version releases often. The `.rubocop.yml` file in the `msgpack` gem and the open-ended dependency definition in the `Gemfile` poses problems for developers building with newer versions of Rubocop and `msgpack`.

Applications with `msgpack` as it currently stands, if they don't explicitly exclude `vendor/bundle/**/*` in their `AllCops` exclusions will experience the following error:

```
Error: RuboCop found unsupported Ruby version 2.3 in `TargetRubyVersion` parameter (in vendor/bundle/ruby/2.5.0/gems/msgpack-1.3.3/.rubocop.yml). 2.3-compatible analysis was dropped after version 0.82.
Supported versions: 2.4, 2.5, 2.6, 2.7
```

This fix does two things:

* Update the `TargetedRubyVersion` to the minimum supported version, which is 2.4
* Restrict the rubocop version in the Gemfile to the current version, so no unintended upgrades are performed that introduce breaking changes without realizing it. This is the recommendation in [Rubocop's installation guide](https://github.com/rubocop-hq/rubocop/blob/master/README.md).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,7 +6,7 @@
 # versions of RuboCop, may require this file to be generated again.
 
 AllCops:
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.4
 
 # Offense count: 3
 Lint/AmbiguousOperator:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,19 +15,17 @@ before_install:
 # http://rubies.travis-ci.org/
 matrix:
   include:
-    - rvm: 2.3.8
-      os: linux
     - rvm: 2.4.5
       os: linux
-    - rvm: 2.5.3
+    - rvm: 2.5.8
       os: linux
-    - rvm: 2.6.1
+    - rvm: 2.6.6
       os: linux
-    - rvm: 2.6.1
+    - rvm: 2.6.6
       os: osx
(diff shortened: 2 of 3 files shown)


tagomoris:
Every time hearing about Rubocop, I want to stop using Rubocop...

Anyway, I'm not against to merging this change, but please make sure CI will pass.

benhayehudi:
> Every time hearing about Rubocop, I want to stop using Rubocop...
I feel your pain!
> Anyway, I'm not against to merging this change, but please make sure CI will pass.
I checked the Travis log and it looks like the build is failing because your Travis config is set to check against different Ruby versions, which makes sense, but it's also checking for v2.3, which is no longer actively supported. Would it be possible to remove that in your Travis config?

tagomoris:
> Would it be possible to remove that in your Travis config?

Sure! I know Ruby 2.3 is too old and it's the result of my laziness. Feel free to remove it (and add newer versions) to pass the CI.

tagomoris:
LGTM. Thanks!

benhayehudi:
My pleasure, thanks for being so responsive!

PR #189, opened 2020-07-15
[Rubocop] 51 files inspected, 1646 offenses detected

mariozaizar:
### Changes

1. Disabled `Style/MutableConstant` on `buffer_spec.rb` because it would raise:
```
FrozenError: can't modify frozen Hash: {}
```
2. Disabled `Lint/UnifiedInteger` on `core_ext.rb` because it would raise:
```
ArgumentError: superclass #<yardoc class Integer> cannot be the same as the declared class #<yardoc class Integer>
```
3. Used Rubocop safe auto-correct:
```ruby
rubocop --safe-auto-correct
Inspecting 51 files
CCCCCCCCCCCWCCCCCCCCCWCCCCCCCCCCCCCCCCWCCCCCCCCCCCC

Offenses:

...

51 files inspected, 1645 offenses detected, 1278 offenses corrected
```
4. Disabled commonly disabled cops with several incidents:
```
Layout/LineLength: # Line is too long
  Enabled: false

Metrics/BlockLength: # Block has too many lines.
  Enabled: false

Metrics/MethodLength: # Method has too many lines.
  Enabled: false

Naming/MethodParameterName: # Method parameter must be at least 3 characters long.
  Enabled: false

Style/FrozenStringLiteralComment: # Missing frozen string literal comment
  Enabled: false
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -34,3 +34,18 @@ Lint/UnusedMethodArgument:
 # Offense count: 128
 Lint/Void:
   Enabled: false
+
+Layout/LineLength: # Line is too long
+  Enabled: false
+
+Metrics/BlockLength: # Block has too many lines.
+  Enabled: false
+
+Metrics/MethodLength: # Method has too many lines.
+  Enabled: false
+
+Naming/MethodParameterName: # Method parameter must be at least 3 characters long.
+  Enabled: false
+
+Style/FrozenStringLiteralComment: # Missing frozen string literal comment
+  Enabled: false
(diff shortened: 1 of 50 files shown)


mariozaizar:
My original intention was to fix as many offenses as I could, but considering some of the `--autocorrect` changes broke some specs, I decided to take another direction and only use `--safe-autocorrect` mode and only ignore a few cops only.

Since I'm not familiar with the internals I won't try to do more than that. In the end, I understand rubocop styling could be considered too opinionated for some, so feel free to close this if the changes are non desired.

Cheers!

PR #190, opened 2020-09-04
Ensure that string hash keys are deduplicated

casperisfine:
@tagomoris, I just saw your Ruby ticket: https://bugs.ruby-lang.org/issues/17147?next_issue_id=16150&prev_issue_id=16651

I wanted to improve msgpack allocations for a while, but I was waiting for the `fstring_*` functions to be expose to C-exts: https://bugs.ruby-lang.org/issues/13381. This is supposed to happen for MRI 3.0, but isn't quite done yet as there is debate on the naming.

In the meantime I think there are a few things that could be done. First modern rubies deduplicate hash keys automatically, but there are a few corner cases. For instance on 2.7 `hash["foo"] = 1` is deduplicated, but not `hash["foo".freeze] = 1`. I suspect it's an oversight.

So I think it would make sense to always deduplicate hash keys.

As a followup I'd like to propose a new `freeze: true` option to deduplicate all strings like Psych: https://github.com/ruby/psych/pull/414

### DIFF
--- a/ext/msgpack/extconf.rb
+++ b/ext/msgpack/extconf.rb
@@ -25,6 +25,43 @@ if defined?(RUBY_ENGINE) && RUBY_ENGINE == 'rbx'
   $CFLAGS << %[ -DDISABLE_RMEM]
 end
 
+# checking if String#-@ (str_uminus) dedupes
+begin
+  a = -(%w(t e s t).join)
+  b = -(%w(t e s t).join)
+  if a.equal?(b)
+    $CFLAGS << ' -DSTR_UMINUS_DEDUPE=1 '
+  else
+    $CFLAGS << ' -DSTR_UMINUS_DEDUPE=0 '
+  end
+rescue NoMethodError
+  $CFLAGS << ' -DSTR_UMINUS_DEDUPE=0 '
+end
+
+# checking if Hash#[]= (rb_hash_aset) dedupes string keys
+h = {}
+x = {}
+r = rand.to_s
+h[%W(#{r}).join('')] = :foo
+x[%W(#{r}).join('')] = :foo
+if x.keys[0].equal?(h.keys[0])
+  $CFLAGS << ' -DHASH_ASET_DEDUPE=1 '
+else
+  $CFLAGS << ' -DHASH_ASET_DEDUPE=0 '
+end
+
+# checking if Hash#[]= (rb_hash_aset) dedupes frozen string keys
+h = {}
+x = {}
+r = rand.to_s
+h[%W(#{r}).join('').freeze] = :foo
+x[%W(#{r}).join('').freeze] = :foo
+if x.keys[0].equal?(h.keys[0])
+  $CFLAGS << ' -DHASH_ASET_DEDUPE_FROZEN=1 '
+else
+  $CFLAGS << ' -DHASH_ASET_DEDUPE_FROZEN=0 '
+end
+
 if warnflags = CONFIG['warnflags']
   warnflags.slice!(/ -Wdeclaration-after-statement/)
 end
--- a/ext/msgpack/unpacker.c
+++ b/ext/msgpack/unpacker.c
@@ -28,7 +28,7 @@
 static int RAW_TYPE_STRING = 256;
 static int RAW_TYPE_BINARY = 257;
 
-static ID s_call;
+static ID s_call, s_uminus;
 
 #ifdef UNPACKER_STACK_RMEM
 static msgpack_rmem_t s_stack_rmem;
@@ -41,6 +41,7 @@ void msgpack_unpacker_static_init()
 #endif
 
     s_call = rb_intern("call");
+    s_uminus = rb_intern("-@");
 }
 
 void msgpack_unpacker_static_destroy()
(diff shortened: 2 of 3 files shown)


casperisfine:
I'm having trouble making it work on 2.4/2.5, but I need to run, I'll put more work into it a bit later.

casperisfine:
It's weird, I can't reproduce the 2.4.x failure on my machine :/ 

But either way this PR is mostly to open a discussion on the kind of string deduplication msgpack could perform.

tagomoris:
Thank you for the proposal!
I also have an idea to introduce an option to apply `String#-@` on map keys to get frozen strings. Of course, It's best if it uses `fstring_*` when msgpack-ruby instantiates strings (if we could use it!).

But we have some points to be cared. We should not change the default behavior. The keys should be frozen (and deduplicated) only when the new option is specified.
In my opinion, `freeze_keys: true` looks nice because it's consistent with `symolize_keys` option. WDYT?

casperisfine:
> We should not change the default behavior.

I'm fine with that, however please note that on 2.6+, string keys are deduplicated by default in `rb_hash_aset`: https://github.com/ruby/ruby/blob/17a27060a75f2b2f773e2ebef56cf943bf1b4e4e/hash.c#L2876-L2894 and message pack is defeating that optimization by pre-freezing the keys. I think that we should fix that at least. If you don't mind I'd like to PR that first before introducing any new option.

Then for older rubies, it can be behind an option, as yes it would be a bit slower.

> In my opinion, freeze_keys: true looks nice because it's consistent with symolize_keys option. WDYT?

I think it's a weird name because string hash keys are always frozen in Ruby. `freeze` is only a side effect of the deduplication.

It makes sense when asking to deduplicate all strings to have a general `freeze: true` option, as it helps explain the consequence, but for keys not so much.

tagomoris:
After thinking from #191, now I'm changing my idea. Map keys are frozen anyway (either by msgpack or Ruby), and the difference is keys are deduplicated or not. User-level matters are frozen or not. Deduplicated or not doesn't matter for users.
So, calling deduplication in default could be no problem.

The only difference is the performance. Calling `String#-@` requires 1.3x times (very roughly, in Ruby 3.0):
```
MBA:~ tagomoris$ time ruby -e '1_000_000.times{ s = String.new("a"); x = -s }'

real	0m0.626s
user	0m0.472s
sys	0m0.055s
MBA:~ tagomoris$ time ruby -e '1_000_000.times{ s = String.new("a") }'

real	0m0.487s
user	0m0.358s
sys	0m0.056s
```

Hmm... 🤔

casperisfine:
> The only difference is the performance. Calling `String#-@` requires 1.3x times (very roughly, in Ruby 3.0):

Indeed, it's not free. So maybe we can just have #191 that does it for "free" for ruby 2.6+, and then when 3.0 land the C api to get a deduplicated string from a `char*`, we can integrate it.

IMHO it's good enough.

What I'd like to propose though is a general `freeze: true` option similar to the one I added to Pysch. It will be slower, but when you know you'll keep that data around and are happy with a CPU vs Memory tradeoof, it would be very useful.

What do you think?

tagomoris:
@casperisfine Honestly speaking, I don't have a concrete idea about how much diffs we'll have about `freeze: true`.
The idea looks good, and it should be great especially when we'll unpack objects for configurations or master data.
On the other hand, I have a question about how many people are using msgpack for such data... Master data etc would be a target use-case.

So, I can't say "yes" or "no" right now. If we can implement that option reasonably, that's good. Otherwise, need more consideration.
@casperisfine Is this a reasonable response to you? It's best to have a concrete use-case (helpful case) if you have it when opening the different issue/pull-req.

casperisfine:
> Is this a reasonable response to you?

Is it perfectly reasonable to me to ask for a use case. I should have started with it in the first place. So here we go.

Our main application loads several hundred MB of static data during boot (what you seem to call Master data). A good par tis I18n translation strings, and various "database" of regional data such as currencies, tax rates etc. These datasets contains quite a lot of duplicated strings, e.g. column names, or even tags etc etc.

So to reduce the memory footprint of our application, we go over all this data right after we load it, and deduplicate what we can, see: https://github.com/Shopify/dedup/blob/9414479a0332220b129d2ce7fe2530906a9ad00a/lib/dedup.rb#L14-L33

The problem is that it's quite slow, just the deduplication is close to 10% of our boot time (down to ~3% with the native c-ext). So we'd like to have the msgpack parser do it, as it would save us from having to walk down the datastructures.

I wanted to wait for `rb_fstring_*()` to be exposed to propose this feature, but I've been waiting for super long now.

As for how many people would use it, well, sure it would be kind of a niche feature, but [since Bootsnap patches `YAML.load_file` to automatically translate YAML is msgpack](https://github.com/Shopify/bootsnap/blob/e8aad0ce200e05c935e0952aadc2e8a5e95e13f4/lib/bootsnap/compile_cache/yaml.rb) some people might end up using it without noticing.

Also once `rb_string_*` is exposed, `freeze: true` might end up faster than `freeze: false` so that might entice people to use it even for runtime parsing.

Let me know if parts of this wasn't clear.

tagomoris:
OK, that use-case is perfectly reasonable.
@casperisfine Could you create a different pull-req to introduce `freeze: true` when you are ready?
I want to make it separated from this context for less confusion.

casperisfine:
> Could you create a different pull-req to introduce `freeze: true` when you are ready?

Absolutely, I wasn't planning to push to this one anymore. Honestly since #191 was merged, I don't think this one adds that much value TBH.

tagomoris:
@casperisfine  Can I close this? (Or could you close this by yourself?)

casperisfine:
Yes, I should have closed it a while ago. Thanks!

PR #191, opened 2020-09-07
Do not pre-freeze string keys if Ruby will dedup them

casperisfine:
Extracted out of https://github.com/msgpack/msgpack-ruby/pull/190

As discussed there, since 2.6 `Hash#[]=` / `rb_hash_aset` automatically deduplicate string keys, unless (and I think it's a bug) they are already frozen. [Source](https://github.com/ruby/ruby/blob/17a27060a75f2b2f773e2ebef56cf943bf1b4e4e/hash.c#L2876-L2894).

So I this case I don't think pre-freezing the keys is as much of an optimization as it was pre 2.6.

### DIFF
--- a/ext/msgpack/extconf.rb
+++ b/ext/msgpack/extconf.rb
@@ -25,6 +25,18 @@ if defined?(RUBY_ENGINE) && RUBY_ENGINE == 'rbx'
   $CFLAGS << %[ -DDISABLE_RMEM]
 end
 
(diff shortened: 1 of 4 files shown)


tagomoris:
This change's intention is quite clear and there's no reason to deny this.
(Note for myself: If Ruby fixes itself to dedup strings even when the key is frozen, those frozen keys will be deduplicated anyway. No problem.)

PR #194, opened 2020-09-09
Implement a freeze: unpacker option

casperisfine:
As discussed in https://github.com/msgpack/msgpack-ruby/pull/190

If set to true all parsed objects will be immediately frozen, and strings will be deduplicated if the Ruby implementation allows it.

This feature is meant to be similar to Pysch's `freeze:` option: https://github.com/ruby/psych/pull/414

For now this option is only useful if you plan to keep the parsed objects around, and want to reduce their memory footprint.

But hopefully Ruby 3.0 should soon expose `rb_fstring_new(const char *ptr, long len)` (likely under another name) which should allow to reduce allocations by directly looking up existing strings without allocating a new one.

cc @tagomoris 

NB: I presume that a java implementation is needed, I'm not too well versed into Java by I can try my hand at it. However I'd rather have your feedback on the MRI implementation first.

### DIFF
--- a/ext/java/org/msgpack/jruby/Decoder.java
+++ b/ext/java/org/msgpack/jruby/Decoder.java
@@ -38,36 +38,38 @@ public class Decoder implements Iterator<IRubyObject> {
   private ExtensionRegistry registry;
   private ByteBuffer buffer;
   private boolean symbolizeKeys;
+  private boolean freeze;
   private boolean allowUnknownExt;
 
   public Decoder(Ruby runtime) {
-    this(runtime, null, new byte[] {}, 0, 0, false, false);
+    this(runtime, null, new byte[] {}, 0, 0, false, false, false);
   }
 
   public Decoder(Ruby runtime, ExtensionRegistry registry) {
-    this(runtime, registry, new byte[] {}, 0, 0, false, false);
+    this(runtime, registry, new byte[] {}, 0, 0, false, false, false);
   }
 
   public Decoder(Ruby runtime, byte[] bytes) {
-    this(runtime, null, bytes, 0, bytes.length, false, false);
+    this(runtime, null, bytes, 0, bytes.length, false, false, false);
   }
 
   public Decoder(Ruby runtime, ExtensionRegistry registry, byte[] bytes) {
-    this(runtime, registry, bytes, 0, bytes.length, false, false);
+    this(runtime, registry, bytes, 0, bytes.length, false, false, false);
   }
 
-  public Decoder(Ruby runtime, ExtensionRegistry registry, byte[] bytes, boolean symbolizeKeys, boolean allowUnknownExt) {
-    this(runtime, registry, bytes, 0, bytes.length, symbolizeKeys, allowUnknownExt);
+  public Decoder(Ruby runtime, ExtensionRegistry registry, byte[] bytes, boolean symbolizeKeys, boolean freeze, boolean allowUnknownExt) {
+    this(runtime, registry, bytes, 0, bytes.length, symbolizeKeys, freeze, allowUnknownExt);
   }
 
   public Decoder(Ruby runtime, ExtensionRegistry registry, byte[] bytes, int offset, int length) {
-    this(runtime, registry, bytes, offset, length, false, false);
+    this(runtime, registry, bytes, offset, length, false, false, false);
   }
 
-  public Decoder(Ruby runtime, ExtensionRegistry registry, byte[] bytes, int offset, int length, boolean symbolizeKeys, boolean allowUnknownExt) {
+  public Decoder(Ruby runtime, ExtensionRegistry registry, byte[] bytes, int offset, int length, boolean symbolizeKeys, boolean freeze, boolean allowUnknownExt) {
     this.runtime = runtime;
     this.registry = registry;
     this.symbolizeKeys = symbolizeKeys;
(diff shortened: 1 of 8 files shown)


casperisfine:
I went ahead and implemented the jruby version.

spec/unpacker_spec.rb
```diff
@@ -662,6 +664,69 @@ def flatten(struct, results = [])
       end
     end
 
+    context 'freeze' do
```
tagomoris, line 667: Could you add a test for `MessagePack.unpack()` ? We can pass options for Unpackers like `MessagePack.unpack(binary, symbolize_keys: true)`

casperisfine: Sure.

casperisfine: Done.

ext/java/org/msgpack/jruby/Decoder.java
```diff
@@ -118,7 +120,12 @@ private IRubyObject consumeUnsignedLong() {
   private IRubyObject consumeString(int size, Encoding encoding) {
     byte[] bytes = readBytes(size);
     ByteList byteList = new ByteList(bytes, encoding);
-    return runtime.newString(byteList);
+    RubyString string = runtime.newString(byteList);
+    if (this.freeze) {
+      string.setFrozen(true);
```
tagomoris, line 125: `Ruby#freezeAndDedupString` looks to freeze&dedup the string, not only dedup.
Is there any reason to call `setFrozen(true)` explicitly here?

tagomoris (approved):
Most things look good. Great!
I left a minor comment about how to get frozen strings in JRuby. Could you check that?

ext/java/org/msgpack/jruby/Decoder.java
casperisfine: Yes, today it's a bit useless, however in theory if you call it with a frozen string, it could directly intern that string without duplicating it first. That's what MRI 3.0 now does, see: https://github.com/ruby/spec/pull/788

I do hope to change that behavior in JRuby at some point, so I figured we might as well call it with a frozen string directly.

ext/msgpack/unpacker.c
```diff
@@ -160,6 +179,16 @@ static inline int object_complete_binary(msgpack_unpacker_t* uk, VALUE str)
 #ifdef COMPAT_HAVE_ENCODING
     ENCODING_SET(str, msgpack_rb_encindex_ascii8bit);
 #endif
+
+#if STR_UMINUS_DEDUPE
+    if(uk->freeze) {
+# if STR_UMINUS_DEDUPE_FROZEN
+        rb_str_freeze(str);
```
casperisfine, line 186: Please don't merge just yet, this part is not quite correct. I'll push a fix in a couple minutes.

casperisfine: It's fixed now.

tagomoris:
Plz ping me when you solve the problem (and failing CI).

casperisfine:
> Plz ping me when you solve the problem (and failing CI).

It should be good now. CI is green. Sorry for the last change.

ext/msgpack/unpacker.c
tagomoris: I can't see the commit to fix this (but the whole commit forcibly pushed).
Could you explain what was wrong and fixed? I want to understand it.

casperisfine: Of course. So the `String#-@` behavior changed a lot on every single version.

Up to 2.5, it had a simple guard statement `return self if frozen?`. Note that frozen doesn't mean "interned". So if you want to deduplicate a string in 2.5 or older, you have to supply a mutable string.

2.6 and 2.7 are a bit smarter, and will deduplicate a supplied string even if it is frozen, however the code is a bit like: `str = str.dup if str.frozen?`, meaning that calling `-@` on a frozen string doesn't give you anything compared to a mutable one.

Then starting in 2.8 (named 3.0 now), if the string is frozen it can directly be put into the interned string table, so freezing the string before calling `-@` will save you an allocation per unique string.

Now my feature checking condition I pushed was a bit broken. The intent was to test wether frozen string could be directly promoted to interned string.

Here's the first version of my test:

```ruby
  a = -(%w(t e s t).join.freeze)
  b = -(%w(t e s t).join.freeze)
  if a.equal?(b)
```

This check was buggy because on 2.6 and 2.7, both expression would return a third instance that was previously created an interned. No huge deal, but it mean we'd do one useless `rb_str_freeze` call for each string.

So I rewrote it as:

```ruby
  s = rand.to_s.freeze
  if (-s).equal?(s)
```

Which did properly work on 2.6+, however it was broken on 2.5- because while the returned string was the same instance, it wasn't interned at all (because `return self if frozen?`).

So I finally pushed:

```ruby
  s = rand.to_s.freeze
  if (-s).equal?(s) && (-s.dup).equal?(s)
```

Which properly check that frozen strings can be promoted to interned strings.

I hope I explained this properly.

tagomoris: That makes sense, perfectly. Thanks!

tagomoris:
I'll merge this change tomorrow if there's no further commits/comments.
The release will be issued after merging #193

casperisfine:
Works for me, thanks!

Issue #192, opened 2020-09-08
Stop Releasing fat gem


tagomoris:
"Fat gem" means gem releases with binaries for each environment. MessegePack-ruby releases are fat gem for Windows environments. (And JRuby release is a kind of fat gem, but it's out of scope of this issue.)

Recent RubyInstaller for Windows includes dev environments (for Ruby 2.4+), and most Ruby users are using that environment. Ruby 2.3 is already EoL, and msgpack-ruby also doesn't support Ruby 2.3.
So, it may be the time to stop releasing fat gem.

See also: https://slide.rabbit-shocker.org/authors/kou/rubykaigi-takeout-2020/

How to check the installation:
```
$ gem i msgpack -v '1.4.0.pre1'
$ ruby -rmsgpack -e 'p MessagePack::VERSION'
```

tagomoris:
https://rubygems.org/gems/msgpack/versions/1.4.0.pre1

emorima:
I've checked this on Windows10 with Ruby 2.7.1. and successed. :)

```
ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x64-mingw32]

C:\Users\emorima>ruby -v
ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x64-mingw32]

C:\Users\emorima>gem i msgpack -v '1.4.0.pre1'
Fetching msgpack-1.4.0.pre1.gem
Temporarily enhancing PATH for MSYS/MINGW...
Building native extensions. This could take a while...
Successfully installed msgpack-1.4.0.pre1
Parsing documentation for msgpack-1.4.0.pre1
Installing ri documentation for msgpack-1.4.0.pre1
Done installing documentation for msgpack after 1 seconds
1 gem installed

C:\Users\emorima>ruby -rmsgpack -e 'p MessagePack::VERSION'
"1.4.0.pre1"
```

tagomoris:
The version `1.4.0.pre1` also works for my M1 mac + Ruby 3.0.0:
```
tagomoris@Moris-Macbook-Air ~ % ruby -v
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [arm64-darwin20]
tagomoris@Moris-Macbook-Air ~ % gem i msgpack -v '1.4.0.pre1'
Fetching msgpack-1.4.0.pre1.gem
Building native extensions. This could take a while...
Successfully installed msgpack-1.4.0.pre1
[...]

PR #196, opened 2020-11-30
Use rb_enc_interned_str when available

casperisfine:
I just got https://github.com/ruby/ruby/pull/3786 merged, which mean in the upcoming Ruby 3.0, C-extensions will be able to return already interned strings with 0 allocations.

For now I use it only in `msgpack_buffer_read_top_as_string` because since we have access to the full C string, it's fairly easy to integrate. However to use it in `read_raw_body_cont` we'd need to do the string buffering in C, I'm open to try but I figured a simpler PR would be easier to review, and also that larger strings are less likely to benefit from this.

@tagomoris what do you think ?

For reference here's the `json` version of this PR: https://github.com/flori/json/pull/451

cc @XrXr @tenderlove @rafaelfranca


ext/msgpack/unpacker.c
```diff
@@ -583,7 +525,7 @@ static int read_primitive(msgpack_unpacker_t* uk)
                 READ_CAST_BLOCK_OR_RETURN_EOF(cb, uk, 4);
                 uint32_t count = _msgpack_be32(cb->u32);
                 if(count == 0) {
-                    return object_complete_string(uk, rb_str_buf_new(0));
+                    return object_complete(uk, rb_utf8_str_new_static("", 0));
```
casperisfine, line 528: Two things here.

Since I removed `object_complete_string` and `object_complete_binary` we have to use API that set the proper encoding directly.

And also `rb_str_buf_new` is meant for string that will be concatenated to, so they are instantiated with a capacity of 63 bytes (so 103bytes total):

```
#define STR_BUF_MIN_SIZE 63
```

Since here we only want an empty string, it's a bit of a waste.

ext/msgpack/buffer.h
```diff
@@ -451,11 +451,45 @@ static inline VALUE msgpack_buffer_read_top_as_string(msgpack_buffer_t* b, size_
...
+        }
+    }
     _msgpack_buffer_consumed(b, length);
     return result;
-}
 
+#else
 
+    result = rb_str_new(b->read_buffer, length);
+#ifdef COMPAT_HAVE_ENCODING
```
casperisfine, line 472: Not sure what the support matrix is, but if we could get rid of `COMPAT_HAVE_ENCODING` we could more easily directly instantiate the string with `rb_utf8_str_new` which would be preferable.

If this for Ruby 1.8?

casperisfine:
@tagomoris any chance this PR could be considered?

tagomoris:
I'll make sure to check this change this week 👀

casperisfine:
Thank you!

ext/msgpack/buffer.h
tagomoris: Yes. Ruby 1.8 is of course out of the support list of msgpack-ruby.
I strongly believe that we can remove `COMPAT_HAVE_ENCODING`, but it should be out of this pull request.

casperisfine: I can try to do it if you'd like, we could then rebase this PR.

ext/msgpack/unpacker.c
tagomoris: I may be misunderstanding your comment. Do you mean:
* the previous code using `rb_str_buf_new` instantiated 63 bytes larger value
* the newer code will not consume additional 63 bytes
Is that correct?

tagomoris (approved):
Totally, this change looks good to me.

ext/msgpack/buffer.h
tagomoris: That is definitely helpful to me :D

ext/msgpack/unpacker.c
casperisfine: The first one. `rb_str_buf_new(0)` will always instantiate a string with 63 bytes of capacity, so a total of 103 bytes.

The new code will directly right size the string, so saving up to 63 bytes.

casperisfine:
> this change looks good to me.

Awesome. I'll wait on https://github.com/msgpack/msgpack-ruby/pull/198 then, and rebase ?

tagomoris:
Merged #198

tagomoris:
Sorry, I'll do further operations (my) tomorrow.

casperisfine:
> Sorry, I'll do further operations (my) tomorrow.

No problem. Thanks for the reviews!

I just rebased the PR.

tagomoris (approved):
LGTM

tagomoris:
Thank you!
My plan is to release this change with other changes about #192 and confirmation about M1 mac build, but please let me know if you want this change earlier.

casperisfine:
> but please let me know if you want this change earlier.

Thanks for asking. There's no big rush. I can point to the git repo anyway.

Thanks a lot for the merge!

Issue #197, opened 2021-01-12
Remove unnecessary ifdef definitions


tagomoris:
Now msgpack-ruby supports Ruby 2.4 or later, but the source tree is still containing many ifdef definitions for much older Ruby versions.
We should remove those to make it easy to maintain our source code.

tagomoris:
`COMPAT_HAVE_ENCODING` https://github.com/msgpack/msgpack-ruby/pull/196/files#r532515695

PR #198, opened 2021-01-12
Remove COMPAT_HAVE_ENCODING ifdef

casperisfine:
Ref: https://github.com/msgpack/msgpack-ruby/issues/197

As discussed in https://github.com/msgpack/msgpack-ruby/pull/196#discussion_r532515695. That flag is for Ruby 1.8 (and MacRuby) support. Both are so old that we can now assume encoding support.

### DIFF
--- a/ext/msgpack/buffer.c
+++ b/ext/msgpack/buffer.c
@@ -23,11 +23,9 @@
 static ID s_replace;
 #endif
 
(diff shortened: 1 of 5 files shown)


tagomoris (approved):
LGTM!
I'll merge this change after the CI green.

PR #206, opened 2021-02-01
Release v1.4.2

tagomoris:


vytstank:
Doesn't this change warrant at least minor version bump? Ruby version changes could break some stacks

PR #203, opened 2021-01-27
Setup GitHub Actions CI

casperisfine:
Travis is very slow these days because they are about to shutdown the free service. Most projects are migrating to GitHub actions.

This PR is just a quick demo, the config file can be adapted to include the rubies and OS you want.

@tagomoris If you are interested, let me know the support list and I can update the matrix.

PR #209, opened 2021-02-02
Enable GitHub Actions

tagomoris:
The change is originally provided by #203


tagomoris:
The current workflow looks working well. I fixed the workflow for TruffleRuby to succeed always even when tests fail.
@casperisfine Thank you for the base config!

tagomoris:
Ah, and @byroot !

casperisfine:
> @casperisfine Thank you for the base config!

You're welcome!

> Ah, and @byroot !

We're the same person 😉

tagomoris:
> We're the same person 😉
I saw only the account name. Sorry!

PR #210, opened 2021-02-09
Clean outdated feature checks

casperisfine:
All the removed checks are always present on 2.4+.


tagomoris (approved):

tagomoris:
Great! Thank you for the clean-up works.

PR #213, opened 2021-03-02
Delete all macro definitions for compatibility about very old Ruby versions

### LINKED ISSUE #197: Remove unnecessary ifdef definitions

tagomoris:
This change is to remove all "compat" macros for Ruby versions 2.0 or earlier.
Fix #197


tagomoris:
I'll merge this for `v1.5.0` after releasing the latest `v1.4.x` for many recent optimizations.

ext/msgpack/buffer.c
```diff
@@ -332,16 +332,9 @@ void _msgpack_buffer_append_long_string(msgpack_buffer_t* b, VALUE string)
...
             rb_funcall(b->io, b->io_write_all_method, 1, string);
-        } else if(!STR_DUP_LIKELY_DOES_COPY(string)) {
-            VALUE s = rb_str_dup(string);
-            ENCODING_SET(s, msgpack_rb_encindex_ascii8bit);
-            rb_funcall(b->io, b->io_write_all_method, 1, s);
         } else {
             msgpack_buffer_append(b, RSTRING_PTR(string), length);
         }
-    } else if(!STR_DUP_LIKELY_DOES_COPY(string)) {
-        _msgpack_buffer_append_reference(b, string);
```
casperisfine, line 343: Seems like we actually lost a feature here, not sure if it was by accident. But from what I gathered `_msgpack_buffer_append_reference` allowed to do zero copy appends from Ruby strings.

I'll try to restore that feature, as it's very common to deal with String and not IOs.

PR #216, opened 2021-03-16
Wrote specs for marginal values of timestamp ranges

tagomoris:
Through the discussions of https://github.com/msgpack/msgpack/pull/263 and https://github.com/msgpack/msgpack/issues/298, I thought that it's worth describing specs for marginal values explicitly.


tagomoris:
Oops, I miss the JVM's missing unsigned int every time 🐼

PR #217, opened 2021-05-06
[WIP] Add option to enforce strictness on core types

shioyama:
We'd like an option which would allow a factory to enforce the condition that all core types (`String`, `Hash`, `Array`, etc.) are only accepted if the object is an instance of the class and not any subclass of it. This happens to be particularly important with `ActiveSupport::HashWithIndifferentAccess` which by default will be treated like a `Hash`, potentially with unexpected results.

Creating PR on Shopify/msgpack-ruby for now for discussion.


shioyama:
Sorry never mind, wrong repo

PR #219, opened 2021-05-11
Add option to strictly limit serialization of core types

shioyama:
_Note: this includes #220, which should be reviewed/merged first._

First, thanks for the great gem! ❤️ 

This is a proposal for a small API + implementation change. Currently, MessagePack will treat any subclass of a "core" type like `Hash` in the same way as the parent type. This can have unexpected results:

```ruby

hash1 = { a: "b" }.with_indifferent_access

hash2 = factory.load(factory.dump(hash1))

hash1[:a]
#=> "b"

hash2[:a]
#=> nil
```

It's also not possible to override this behaviour by, say, defining an extension type for a core type subclass since such a type will be ignored:

```ruby
factory.register_type(
  0x00,
  ActiveSupport::HashWithIndifferentAccess,
  packer: ->(value) { MessagePack.dump(value) },
  unpacker: ->(value) { MessagePack.load(value).with_indifferent_access }
)
factory.load(factory.dump(hash1)).class
#=> Hash
```

This presents a problem if MessagePack is being used in a context where instances of classes like `ActiveSupport::HashWithIndifferentAccess` may be present.

This PR is a suggestion for how this might be dealt with. I've added a `strict` option to `MessagePack::Factory#initialize` which, when `true`, causes the factory to raise a new exception, `MessagePack::PackError`, if the factory attempts to pack any object which is neither an instance of a core type, nor an instance of an extension type (or subclasses thereof).

Continuing from the above example:

```ruby
factory = MessagePack::Factory.new(strict: true)
#=> #<MessagePack::Factory:0x000055df4342e348>

hash1 = { a: "b" }.with_indifferent_access
factory.dump(hash1)
# raises MessagePack::PackError
```

The change here also (as a byproduct of how it is implemented) makes it possible, with either `strict: true` _or_ `strict: false`, to define an extension type for a subclass of a core type, and have that extension type work.

```ruby
factory = MessagePack::Factory.new(strict: false) # or `strict: true`, doesn't impact the result

factory.register_type(
  0x00,
  ActiveSupport::HashWithIndifferentAccess,
  packer: ->(value) { MessagePack.dump(value) },
  unpacker: ->(value) { MessagePack.load(value).with_indifferent_access }
)
factory.load(factory.dump(hash1)).class
#=> ActiveSupport::HashWithIndifferentAccess
```

I think this is a more natural result.

I have not implemented the jruby code yet since I'd like to discuss first whether this change makes sense, and if so whether the implementation is appropriate. It may also make sense to add a `strict` option to `register_type` (to enforce strictness on extension types as well), but I haven't done that since I wanted to get your thoughts on this first. Happy to update the PR incorporating any feedback, etc.

Thanks!

### DIFF
--- a/doclib/msgpack/error.rb
+++ b/doclib/msgpack/error.rb
@@ -1,5 +1,8 @@
 module MessagePack
 
+  class PackError < StandardError
+  end
+
   class UnpackError < StandardError
   end
 
--- a/ext/java/org/msgpack/jruby/Encoder.java
+++ b/ext/java/org/msgpack/jruby/Encoder.java
@@ -118,7 +118,7 @@ public class Encoder {
       appendInteger((RubyInteger) object);
     } else if (object instanceof RubyFloat) {
       appendFloat((RubyFloat) object);
-    } else if (object instanceof RubyString) {
+    } else if (object.getType() == runtime.getString()) {
       appendString((RubyString) object);
     } else if (object instanceof RubySymbol) {
       if (hasSymbolExtType) {
@@ -126,9 +126,9 @@ public class Encoder {
       } else {
         appendString(((RubySymbol) object).asString());
       }
-    } else if (object instanceof RubyArray) {
+    } else if (object.getType() == runtime.getArray()) {
       appendArray((RubyArray) object);
-    } else if (object instanceof RubyHash) {
+    } else if (object.getType() == runtime.getHash()) {
       appendHash((RubyHash) object);
     } else if (object instanceof ExtensionValue) {
       appendExtensionValue((ExtensionValue) object);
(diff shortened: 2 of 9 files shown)


tagomoris:
The entire idea looks good, but this change seems to contain two different changes to me:
* adding `strict` option to raise an error if a serialized object is not an instance of defined classes (but subclasses)
* changing the current behavior to serialized as a core type object (for example, Hash) if an object is an instance of subclasses even when an extension type is registered for the subclass of core types

In my opinion, it may be ok to say "it's a bug" about the 2nd point. It should be unexpected behavior for users in most cases.
But the first one is a clear new feature.

So, my idea is, it looks better to split those changes into independent pull requests. I want to review bugfix and new feature separately.
@shioyama What do you think?

shioyama:
> it looks better to split those changes into independent pull request.

Definitely, that makes sense. I'll start by creating a separate PR for the bug fix 👍

shioyama:
Due to changes in #219, I'll need to code this change differently, so I'll make a fresh PR and close this one.

PR #220, opened 2021-05-12
Support primitive type extensions

shioyama:
Extracted from #219

Currently MessagePack does not support extensions that subclass "primitive" types like `Hash`. e.g.

```ruby

hash = {}.with_indifferent_access

factory.register_type(
  0x00,
  ActiveSupport::HashWithIndifferentAccess,
  packer: ->(value) { MessagePack.dump(value) },
  unpacker: ->(value) { MessagePack.load(value).with_indifferent_access }
)
factory.load(factory.dump(hash)).class
#=> Hash
```

With this change, this works, and the result is `ActiveSupport::HashWithIndifferentAccess`.

The code change also supports subclasses of `Array` and `String`, since these can also have instances; other primitive types cannot have subclasses that can be instantiated so I've left them unchanged.

### DIFF
--- a/ext/java/org/msgpack/jruby/Encoder.java
+++ b/ext/java/org/msgpack/jruby/Encoder.java
@@ -119,7 +119,9 @@ public class Encoder {
     } else if (object instanceof RubyFloat) {
       appendFloat((RubyFloat) object);
     } else if (object instanceof RubyString) {
-      appendString((RubyString) object);
+      if (object.getType() == runtime.getString() || !tryAppendWithExtTypeLookup(object)) {
+        appendString((RubyString) object);
+      }
     } else if (object instanceof RubySymbol) {
       if (hasSymbolExtType) {
         appendOther(object, destination);
@@ -127,9 +129,13 @@ public class Encoder {
         appendString(((RubySymbol) object).asString());
       }
     } else if (object instanceof RubyArray) {
-      appendArray((RubyArray) object);
+      if (object.getType() == runtime.getArray() || !tryAppendWithExtTypeLookup(object)) {
+        appendArray((RubyArray) object);
+      }
     } else if (object instanceof RubyHash) {
-      appendHash((RubyHash) object);
+      if (object.getType() == runtime.getHash() || !tryAppendWithExtTypeLookup(object)) {
+        appendHash((RubyHash) object);
+      }
     } else if (object instanceof ExtensionValue) {
       appendExtensionValue((ExtensionValue) object);
     } else {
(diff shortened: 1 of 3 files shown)


shioyama:
The jruby tests are failing since I haven't implemented the change in jruby yet. I'll start working on that now.

shioyama:
@tagomoris This is the first "bug fix" PR. I added changes for jruby and tests are passing, and this one I think should be straightforward since the change is small.

よろしくお願いします！ 🙇‍♂️

shioyama:
Ok uncovered an issue with the jruby change, investigating now 👀

tagomoris:
With this updated code, `Packer` will call `to_msgpack` method of subclass (of core types) objects. It can cause unexpected but huge performance impact on high traffic serialization. I can't say it never happens. We can easily imagine cases like "Yes, it's almost a Hash, but I want it to be a subclass of Hash for a utility method".
@shioyama could you consider a way to use the core type serialization functions for subclasses of core types too if ext types are not registered?

shioyama:
> could you consider a way to use the core type serialization functions for subclasses of core types too if ext types are not registered?

Yup, I'll try to do that 👍

shioyama:
@tagomoris I've updated the C code to avoid hitting `to_msgpack` in case no extension exists for a core type subclass. I'm working on the java code now but it will be the same unless there are any issues with this implementation.

tagomoris:
@shioyama The implementation itself looks good, but I got seriously confused about naming. Can we use a name like "try writing with ext type lookup" instead of the one prefixed with `_`?

shioyama:
> Can we use a name like "try writing with ext type lookup" instead of the one prefixed with _?

Yeah definitely 👍 I couldn't think of anything and wasn't really sure about any possible conventions. I'll make the change.

shioyama:
Ok tests are now passing. I'm unsure about the java/C conventions so please let me know if I should rename any methods etc. Also the implementations are slightly different: the C code falls back to the extension type in the `write_string_value`, `write_hash_value` etc methods, whereas in Java the fallback is done in `appendObject`, because this was slightly easier (and anyway it seems there is not a 1-1 relationship between the two).

Thanks for all the feedback!

shioyama:
I made one final change in implementation and squashed everything into two commits (one for failing tests, one for code changes).

The java and C implementation were different: in java I was checking for the hash subclass extension type in the equivalent of the C switch statement (`appendObject` in `Encoder`), whereas in C I was doing this in each of the respective write methods (for Hash, Array and String). I thought about it and find the former preferable, so I've updated the code to do the same in C as well: so `msgpack_packer_write_string_value`, `msgpack_packer_write_hash_value` and `msgpack_packer_write_array_value` remain unchanged, and changes all happen in `msgpack_packer_write_value`.

My thinking for this is that a method like `msgpack_packer_write_hash_value` should just write a Hash value, as its name implies.

tagomoris (approved):
LGTM!
I'll re-check code again and merge this early next week (sorry, I don't have time right now)

shioyama:
Thanks very much!

tagomoris (approved):

PR #221, opened 2021-05-13
Add option to strictly limit serialization of core types

shioyama:
Re-implementation of #219 on top of #220.

This is a small API + implementation change. Currently, MessagePack will treat any subclass of a "core" type like `Hash` in the same way as the parent type. This can have unexpected results:

```ruby

hash1 = { a: "b" }.with_indifferent_access

hash2 = factory.load(factory.dump(hash1))

hash1[:a]
#=> "b"

hash2[:a]
#=> nil
```

This PR adds a `strict` option to `MessagePack::Factory#initialize` which, when `true`, causes the factory to raise a new exception, `MessagePack::PackError`, if the factory attempts to pack any object which is neither an instance of a core type, nor an instance of an extension type (or subclasses thereof).

Continuing from the above example:

```ruby
factory = MessagePack::Factory.new(strict: true)
#=> #<MessagePack::Factory:0x000055df4342e348>

hash1 = { a: "b" }.with_indifferent_access
factory.dump(hash1)
# raises MessagePack::PackError
```


shioyama:
Converted to draft mode since java implementation is not done yet.

shioyama:
@tagomoris This one is passing tests now too, please 👀 when you have a chance (but after #220).

shioyama:
I forgot to implement `MessagePack::Factory#strict_types?` in java, so just added that now.

We actually don't _need_ this method, it was just useful when I was testing the feature. Happy to remove if you think it's better to leave it out.

tagomoris:
I'm positive to including this feature, but I have some concerns and questions on this change:

### the Option name
The option name `strict` seems uncertain... actually, this feature is to prohibit serializing unregistered classes. Can we realize "it is strict"?

### Target classes are limited
This change raises PackError only for subclasses of Hash, Array, and String. How about other classes like Integer, Float, TrueClass/FalseClass, and NilClass?

### Classes not registered but with `to_msgpack`
The current implementation will raise PackError for objects of classes that are not core types nor registered as ext types, but have a method `to_msgpack`. It is an incompatibility. Is it a right thing?

shioyama:
Thanks for the feedback!

> The option name strict seems uncertain... actually, this feature is to prohibit serializing unregistered classes. Can we realize "it is strict"?

I'm also not sure what is best here, since `strict` only applies to core/primitive types and not extension types. If we want to _really_ be strict then we would need to apply strictness to extension types as well, but also with some way to disable that (in our case we actually only want strictness on the core types, not extension types...) So this is where I am hesitating.

Let me think a bit more about this and propose some ideas for how we could do this. If you have suggestions please let me know!

> This change raises PackError only for subclasses of Hash, Array, and String. How about other classes like Integer, Float, TrueClass/FalseClass, and NilClass?

I don't believe those other classes can have instantiatable subclasses, so I left them out. e.g.

```ruby
class Value < FalseClass
end  
Value.new
# raises NoMethodError: undefined method `new' for Value:Class
```

Is there any scenario where this or other other core types can be subclassed and instantiated?

> The current implementation will raise PackError for objects of classes that are not core types nor registered as ext types, but have a method to_msgpack. It is an incompatibility. Is it a right thing?

You're right, this is a code path I hadn't considered. I can change this so it only raises `PackError` if the object has no `to_msgpack` method, would that make sense? I'll also add a test for this.

tagomoris:
About the naming, I don't have any fascinating solutions too. Names like "serialize-exact-classes-only" can be an option, but it's also an incorrect name because the current implementation doesn't raise errors for subclasses of classes registered for ext types.

Yes, this is another concern: The current implementation doesn't touch ext type lookups. Is it intentional? Shouldn't we affect this option on ext type lookups? (This depends on naming...)

> I don't believe those other classes can have instantiatable subclasses, so I left them out.

Oh, that's true. I want a comment about this fact in the change for the time when I revisit the code in the future.

> You're right, this is a code path I hadn't considered. I can change this so it only raises PackError if the object has no to_msgpack method, would that make sense?

This totally makes sense to me.

shioyama:
>  in our case we actually only want strictness on the core types, not extension types...

Actually I realized this is not really true: we do need control over extension type subclasses, and I would imagine that this would be a common requirement.

I'm thinking of an interface like this:

```ruby
factory = MessagePack::Factory.new(match_subclasses: false)

factory.register_type(0x00, Foo, packer: ..., unpacker: ..., match_subclasses: true)
factory.register_type(0x01, Bar, packer: ..., unpacker: ..., match_subclasses: false)
```

The name `match_subclasses` could be something else, but the idea is that the default of `register_type` defaults to the default of the factory, so if you pass `match_subclasses: false` to `MessagePack::Factory.new`, then `register_type` will also default to not match subclasses of extention types. But you can always override this in `register_type`.

I'm not sure about the name (`match_subclasses`? `match_descendants`? `strict_types`?...) but I think this general convention makes sense and allows full control over how types are applied (both core types and extension types).

WDYT?

tagomoris:
Using a common rule for both core types and ext types sounds nice to me.

Yes, "a common rule", so I think it may be too much to have an additional option only for registering ext types. Users can create another Factory (with different option value) anytime, or also can register subclasses themselves in addition to the base class (ah, it should be a bad idea because it needs a different type id, though).
And the additional option should make implementation very complex. This idea (additional option for registering ext types) seems so unbalanced to me.

For the naming, I personally prefer having options with default `false` because it makes less confusion for readers (I think). From this viewpoint, `strict_types` looks better than `match_subclasses`.
My preferable name can be ... `strict_class_matching` or `ignore_subclass` ... (still thinking).

shioyama:
> And the additional option should make implementation very complex. This idea (additional option for registering ext types) seems so unbalanced to me.

Yes that's true, and anyway it is possible to raise the same `MessagePack::PackError` from the extension type packer block if the object has a class that does not match, e.g.:

```ruby
factory.register_type(0x00, Foo, packer: ->(value) do
  raise MessagePack::PackError unless value.class == Foo
  # ...
end)
```

So unlike the core types (where there is no way to apply "strictness" currently) with extension types the user can control this behaviour.

> For the naming, I personally prefer having options with default false because it makes less confusion for readers (I think). From this viewpoint, strict_types looks better than match_subclasses.

👍 

I wonder if the name should mention "core" or "primitive" in it, something to indicate it only applies to the core types 🤔 

I'll keep thinking too.

shioyama:
>  I can change this so it only raises PackError if the object has no to_msgpack method, would that make sense?

Oh wait, there's a problem here: subclasses of core types inherit `to_msgpack`, so according to this change we would never raise on them.

I see two solutions:
1. special case for core subtypes: they always raise `MessagePack::PackError` if `strict_types` is enabled and no extension type is defined for them, regardless of presence of `to_msgpack`
2. keep code as-is and maintain consistency: `strict: true` means that `to_msgpack` is ignored when packing

I don't see any other solutions to this problem... and I don't really like 1., which is why I originally coded it this way (2).

shioyama:
Actually, I think `to_msgpack` is not really compatible with the idea of "strict class matching", since by definition it is an inherited property of a type. I think we can just make it clear that `strict: true` (or whatever we name it) overrides `to_msgpack`. It's a new thing so existing code would still work. What do you think?

shioyama:
There is one other solution: modify `to_msgpack` defined on core types to raise `MessagePack::PackError` if the packer has strict types enabled. This way, enabling the strict flag on the packer would work as expected (raise on core types even though they have a `to_msgpack` defined), and would be overrideable on core type subclasses by redefining `to_msgpack` on the subclass to not raise.

I'll try this.

tagomoris:
We can check whether an object has an intentionally implemented `to_msgpack` by `obj.class.method_defined?(:to_msgpack, inherit: false)` if we want to do this check strictly.

shioyama:
> We can check whether an object has an intentionally implemented `to_msgpack` by `obj.class.method_defined?(:to_msgpack, inherit: false)` if we want to do this check strictly.

Ah ok, that makes sense and follows the pattern nicely 👍 Thanks for the quick feedback!

shioyama:
OK so I think there is already an inconsistency in how `to_msgpack` is handled. I've created a separate PR to discuss: https://github.com/msgpack/msgpack-ruby/pull/222

shioyama:
It doesn't seem that `method_defined?` accepts an `inherited` option 🤔  Seems like the only option is to actually check `instance_methods(false)`.

shioyama:
Ok it's working on mruby versions >= 2.7, but I'll need to use `instance_methods.include?(:to_msgpack, false)` in order to make it work in earlier versions I guess...

shioyama:
Ok calling `instance_methods(false).include?(:to_msgpack)` works, I'll just need to do the same in java.

Wondering how much of a perf hit this will incur. It's a pretty narrow corner case.

shioyama:
@tagomoris I dug through other implementations of msgpack and discovered there is an identically named `strict_types` option in [msgpack-python]:

https://github.com/msgpack/msgpack-python/pull/158

Description of the option:

> If set to true, types will be checked to be exact. Derived classes
> from serializeable types will not be serialized and will be
> treated as unsupported type and forwarded to default.
 > Additionally tuples will not be serialized as lists.
 > This is useful when trying to implement accurate serialization
 > for python types.

Since this seems to be a common pattern, I've made the change to stick with the name `strict_types` here as well, so instead of `MessagePack::Factory.new(strict: true)` it would be `MessagePack::Factory.new(strict_types: true)`. 

I have also removed the `#strict_types?` method since I don't think it is really needed.

What do you think? If this all looks good I'll squash commits and clean things up.

shioyama:
Looking more closely at the implementation in Python, there are two key differences:

1. the `strict_types` option is on the packer, not the factory (there is no concept of factory in msgpack-python)
2. `strict_types` applies to extension types as well as core types (we decided not to do this)
3. the "bug" fixed in https://github.com/msgpack/msgpack-ruby/pull/220 is not fixed, meaning that with `strict_types` `false` an extension for a subclass of core types is ignored.

I don't think there is any need to follow conventions in other languages, but just for reference I thought I'd point this out.

doclib/msgpack/error.rb
```diff
@@ -1,5 +1,8 @@
 module MessagePack
 
+  class PackError < StandardError
+  end
```
tagomoris, line 4: This class doesn't have `attr_accessor :error_value`. Is it intentional?

tagomoris:
I want to review this change again after rebasing on the current master... anyway, why did you add `strict_types` option on `Factory`, not on `Packer`? That option is to control the packer's behavior, so it seems ok to have that option on `Packer#initialize` (and its option can be specified on `Factory#packer` too).

tagomoris:
> strict_types applies to extension types as well as core types (we decided not to do this)

Oh, did you decide not to do that? In my opinion, users should expect that the option should be applied on ext types too (unless the option name contains "core types" or "primitive types"). I prefer to enable the same rule on core types and ext types when we enable this option.

shioyama:
> I want to review this change again after rebasing on the current master...

Done 👍 

> why did you add `strict_types` option on `Factory`, not on `Packer`? That option is to control the packer's behavior, so it seems ok to have that option on `Packer#initialize` (and its option can be specified on `Factory#packer` too).

Sure I can do that 👍 

> Oh, did you decide not to do that? In my opinion, users should expect that the option should be applied on ext types too (unless the option name contains "core types" or "primitive types"). I prefer to enable the same rule on core types and ext types when we enable this option.

Yes, I agree this is more intuitive, and I can make this change. The problem is that for our use case, this actually doesn't work: we need to match on subclasses of one extension type, but also apply other types strictly. I'm not sure if other people would have similar requirements. 🤔 

But with #220 now merged, I believe I can achieve the same thing by defining extensions for `Hash`, `Array` and `String` which simply raise an exception, which our code can then catch and handle appropriately. So a `Hash` would be treated the same as before, but any subclass would match the `Hash` extension type and call the packer (which would raise and not allow it to be encoded).

This should work, but it's a bit ugly since we need to reserve three codes that will never be used for each of these types.

So I'm a bit stuck: I feel that you are right that `strict_types` _should_ apply to extensions (like msgpack-python) but doing this means that a valid use case like ours is not actually supported (except with the weird trick I mentioned above).

The only way to really support this particular use case would be to allow each extension type to specify strictness (`register_type(..., strict: true)` or similar). But this adds more complexity.

🤔

doclib/msgpack/error.rb
shioyama: Oh that's strange... not intentional, I'll add it back, thanks!

tagomoris:
@shioyama My quick idea about your use-case is to have an option (in addition to `strict_types`) to ignore `strict_types` option only for ext types (for example, `instrict_ext_types`) with default `false`.  You can specify `strict_types: true, instrict_ext_types: true` but other people can use `strict_types: true` without additional care of ext types.

Another idea is to have a `type_matching_mode` option (instead of `strict_types`) which can receive symbols to specify modes, like `:standard`(matches to subclasses), `:strict`(matches to the specified classes only), `:strict_core_types` (matches to the specified classes only for core types), `:strict_ext_types` (matches to the specified classes only for ext types).

shioyama:
Hi, sorry for the delay in following up. #220 was enough to solve the immediate problem for us, but this is still something that would make things a lot cleaner.

> My quick idea about your use-case is to have an option (in addition to strict_types) to ignore strict_types option only for ext types (for example, `instrict_ext_types`) with default false. You can specify `strict_types: true`, `instrict_ext_types: true` but other people can use `strict_types: true` without additional care of ext types.

The problem is that this is still on/off for all extension types. The interface that would make most sense for us would be to have an option on `register_type` which would enable/disable strictness on a per-extension type basis. This would allow fine-grained control over strictness, with the option defaulting to `false` so existing behaviour is maintained.

An all-or-nothing approach means that we would need to `raise ... unless value.instance_of?(...)` in many extension type packers, which feels wrong.

My suggestion would be to have `strict_types` on the factory/packer, which would be inherited as the default in extension types but would be overridable in `register_type`. So the default `strict_types: false` would be inherited in extension types, but if you set `strict_types: true` then extension types would also apply strictness, but you would always be able to override this with:

```ruby
factory.register_type(0x01, MyClass, strict: false) # would override strict_types: true on the factory
```

What do you think?

tagomoris:
Having `strict` option on `Factory/Packer`'s `#register_type` sounds a consistent way, but I'm afraid that implementation would become very complex.
Now I have an idea to reduce the complexity of the implementation... the new option `strict_class_match` (very long name to avoid using it casually) which does:

```ruby
packer.register_type(0x01, MyClass, strict_class_match: true)
# ↑ this is equal to ↓
packer.register_type(0x01, MyClass) do |obj|
  raise PackError, "type mismatch" unless obj.class == MyClass
  obj.to_msgpack_ext
end
```

This will require additional performance overhead, but it can be implemented very easily and cleanly.

tagomoris:
I'm sorry to say this after the long story on this pull request, but now I'm afraid that this feature might be too complex but not so valuable for actual use-cases. @shioyama can solve their problem by their way:

> But with #220 now merged, I believe I can achieve the same thing by defining extensions for Hash, Array and String which simply raise an exception
https://github.com/msgpack/msgpack-ruby/pull/221#issuecomment-843636358

So, I want to close this if there is no one to support this idea strongly. Thoughts?

PR #222, opened 2021-05-14
Failing test for behavior of to_msgpack without extension type

shioyama:
I'm creating this PR to demonstrate what seems like inconsistent, or at least unclear, behaviour of `to_msgpack` when called from `MessagePack::Factory#dump`.

```ruby
describe '#to_msgpack' do
  it 'does not raise on classes that define public to_msgpack method' do
    stub_const('Foo', Class.new do
      def to_msgpack(*)
        "to_msgpacked"
      end
    end)
    object = Foo.new
    factory = described_class.new

    expect(factory.dump(object)).to eq('to_msgpacked')
  end
end
```

I would expect this test to pass, but it fails. `MessagePack::Factory#dump` here does not raise anything, it simply returns a blank string. Is this the correct behaviour?

### DIFF
--- a/spec/factory_spec.rb
+++ b/spec/factory_spec.rb
@@ -364,4 +364,18 @@ describe MessagePack::Factory do
       expect(MessagePack.unpack(MessagePack.pack(dm2))).to eq(dm2)
     end
   end
+
+  describe '#to_msgpack' do
+    it 'does not raise on classes that define public to_msgpack method' do
+      stub_const('Foo', Class.new do
+        def to_msgpack(*)
+          "to_msgpacked"
+        end
+      end)
+      object = Foo.new
+      factory = described_class.new
+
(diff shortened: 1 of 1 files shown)


shioyama:
The problem is that the test class does not define `to_msgpack_ext`, which would normally fall through to `to_msgpack` for a core type but not for another class.

But I'm still not sure what the behaviour should be in this case. It seems to me that it should be one of:
1. raise some exception
2. return `"to_msgpacked"`

shioyama:
Ok, I see, the problem is that `to_msgpack` should accept a `packer` and write to that:

```ruby
stub_const('Foo', Class.new do
  def to_msgpack(packer)
    packer.write_string("to_msgpacked")
    packer
  end
end)
```

Semantically it feels a bit strange, because `to_msgpack` makes me think the method converts an object to a msgpack string. I'll close this though since it's my misunderstanding of the API.

PR #225, opened 2021-08-26
Optimized symbol parsing

casperisfine:
In our app we parse a lot of msgpack payloads containing symbols, and as a result `String#to_sym` called by MessagePack is one of our biggest hotspots, so we would like to optimize it.

However it shows that if we add special handling for symbol parsing we can do it twice faster.

```ruby
require "benchmark/ips"
require "msgpack"

optimized_factory = MessagePack::Factory.new
optimized_factory.register_type(0x00, Symbol, packer: :to_msgpack_ext, unpacker: MessagePack::OPTIMIZED_SYMBOL_UNPACKER)
MessagePack::DefaultFactory.register_type(0x00, Symbol)

raise "BUG" unless MessagePack.load(MessagePack.dump(:foo)) == :foo
raise "BUG" unless optimized_factory.load(MessagePack.dump(:foo)) == :foo

PAYLOAD = MessagePack.dump(20.times.map { |i| :"foo#{i}" })
regular_unpacker = MessagePack::DefaultFactory.unpacker
optimized_unpacker = optimized_factory.unpacker

Benchmark.ips do |x|
  x.report('regular') do
    regular_unpacker.feed(PAYLOAD).full_unpack
  end
  x.report('optimized') do
    optimized_unpacker.feed(PAYLOAD).full_unpack
  end
  x.compare!
end
```

```bash
             regular     4.342k i/100ms
           optimized     9.156k i/100ms
             regular     43.592k (± 1.2%) i/s -    221.442k in   5.080628s
           optimized     92.098k (± 1.0%) i/s -    466.956k in   5.070682s

Comparison:
           optimized:    92098.2 i/s
             regular:    43592.3 i/s - 2.11x  (± 0.00) slower
```

cc @shioyama 

@tagomoris what do you think?

### DIFF
--- a/ext/msgpack/buffer.h
+++ b/ext/msgpack/buffer.h
@@ -494,4 +494,9 @@ static inline VALUE msgpack_buffer_read_top_as_string(msgpack_buffer_t* b, size_
 #endif // HAVE_RB_ENC_INTERNED_STR
 }
 
+static inline VALUE msgpack_buffer_read_top_as_symbol(msgpack_buffer_t* b, size_t length)
+{
+    return rb_str_intern(msgpack_buffer_read_top_as_string(b, length, true, false));
+}
+
 #endif
--- a/ext/msgpack/factory_class.c
+++ b/ext/msgpack/factory_class.c
@@ -32,6 +32,8 @@ struct msgpack_factory_t {
     msgpack_packer_ext_registry_t pkrg;
     msgpack_unpacker_ext_registry_t ukrg;
     bool has_symbol_ext_type;
+    bool optimized_symbol_ext_type;
+    int symbol_ext_type;
 };
 
 #define FACTORY(from, name) \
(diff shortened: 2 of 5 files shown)


tagomoris:
@casperisfine Honestly, it seems a little too tricky to me that it has a Proc as an "OPTIMIZED" unpacker but it's actually a usual (not-optimized) Proc and it'll not be used actually. This trick will make me confused in the future.
Could you consider an alternative option to opt-in the optimization?

byroot:
> Could you consider an alternative option to opt-in the optimization?

If you don't like this one, then of course, but what do you have in mind?

tagomoris:
Sorry for the late reply on this.

> If you don't like this one, then of course, but what do you have in mind?

I don't have any concrete idea on it, but for example, having a boolean(true/false) keyword arg to enable the optimized symbol unpacker can be an option.
"Optimized" is the feature here, so using a (non-optimized) Proc object on JRuby should be unexpected behavior for users. In my idea, specifying the option to use the un-implemented optimized unpacker should raise an error (on JRuby).

casperisfine:
@tagomoris I made the modification you requested.

tagomoris:
LGTM.
I'll add a change on JRuby to raise an error when `optimized_symbols_parsing: true` is specified before releasing this change.

tagomoris:
Currently, I have difficulties to install/run JRuby on my laptop (because of jffi / native libraries). So, it may take a while.

PR #227, opened 2021-09-30
Fix undefined method `runtime' for JRuby:Module on JRuby 9.3.0.0

hiroyuki-sato:
This PR fix #226. I want to confirm the CI result.

PR #228, opened 2021-09-30
use a new loading mechanism on jruby

ahorek:
this PR fixes https://github.com/msgpack/msgpack-ruby/issues/226 compatibility for JRuby 9.3.x
and drops support for JRuby 9.1.x. Since the [minimum supported Ruby version](https://github.com/msgpack/msgpack-ruby/blob/master/msgpack.gemspec#L23) is 2.4 and JRuby 9.1.x targets 2.3 I guess it's ok.

see also https://github.com/jruby/jruby/issues/6868

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu]
-        ruby: ['jruby']
+        ruby: ['jruby-9.2.19.0', 'jruby-head']
     runs-on: ${{ matrix.os }}-latest
     steps:
       - uses: actions/checkout@v2
--- a/lib/msgpack.rb
+++ b/lib/msgpack.rb
@@ -1,9 +1,8 @@
 require "msgpack/version"
 
 if defined?(RUBY_ENGINE) && RUBY_ENGINE == "jruby" # This is same with `/java/ =~ RUBY_VERSION`
-  require "java"
   require "msgpack/msgpack.jar"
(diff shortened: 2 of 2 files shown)


hiroyuki-sato:
Hello, @ahorek Thank you for creating this PR. 
This PR better than my change(#227). correct?
If so, I'll close my PR. 

Thank you.

ahorek:
yes, thanks for letting us know about the problem. Your change is an option, but the new syntax is preferred, since we don't need compatibility with EOL versions of JRuby.

hiroyuki-sato:
Thank you for your comment. #227 closed.

tagomoris:
Thanks for creating the change!
CI is failing, but it's fixed at #230 
```
  1) MessagePack::Unpacker ensure string hash keys are deduplicated
     Failure/Error: expect(hashes[0].keys.first).to equal(hashes[1].keys.first)

       expected #<String:4020> => "foo"
            got #<String:4024> => "foo"

       Compared using equal?, which compares object identity,
       but expected and actual are not the same object. Use
       `expect(actual).to eq(expected)` if you don't care about
       object identity in this example.
     # ./spec/unpacker_spec.rb:37:in `block in <main>'
```

ivoanjo:
While testing #239 I ran into an issue: the new loading mechanism only works from JRuby 9.2.1.0+ on, aka it doesn't work on the 9.2.0.0 release.

I don't think it's worth fixing it; just thought I drop here a note in case someone runs into 
``NoMethodError: undefined method `load_ext' for JRuby::Util:Module`` on msgpack 1.4.4+ on JRuby 9.2.0.0, which hopefully is absolutely nobody.

PR #229, opened 2021-09-30
fix warnings

ahorek:
* Serializable classes without serializeVersionUID
* Generic classes used as raw types
* Optimize runtime access
* Remove double casts


tagomoris:
I want to see the CI result based on #230 (if it's green, of course)

ahorek:
@tagomoris done!

tagomoris (approved):
LGTM

PR #230, opened 2021-09-30
dedup hash keys

ahorek:
JRuby 9.3 does dedup hash keys since https://github.com/jruby/jruby/commit/454ef8b3cec17a842c08a394d4453ef1dd82f0fa

fixes a test failure
```
  1) MessagePack::Unpacker ensure string hash keys are deduplicated
     Failure/Error: expect(hashes[0].keys.first).to equal(hashes[1].keys.first)

       expected #<String:4020> => "foo"
            got #<String:4024> => "foo"

       Compared using equal?, which compares object identity,
       but expected and actual are not the same object. Use
       `expect(actual).to eq(expected)` if you don't care about
       object identity in this example.
     # ./spec/unpacker_spec.rb:37:in `block in <main>'
```

### DIFF
--- a/ext/java/org/msgpack/jruby/Decoder.java
+++ b/ext/java/org/msgpack/jruby/Decoder.java
@@ -140,9 +140,15 @@ public class Decoder implements Iterator<IRubyObject> {
     RubyHash hash = RubyHash.newHash(runtime);
     for (int i = 0; i < size; i++) {
       IRubyObject key = next();
-      if (this.symbolizeKeys && key instanceof RubyString) {
+      if (key instanceof RubyString) {
+        if (this.symbolizeKeys) {
           key = ((RubyString) key).intern();
+        } else {
+          key.setFrozen(true);
+          key = runtime.freezeAndDedupString((RubyString) key);
+        }
       }
+
       hash.fastASet(key, next());
     }
     return hash;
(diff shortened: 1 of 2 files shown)


tagomoris:
@ahorek Could you rebase this branch on the current master?

ext/java/org/msgpack/jruby/Decoder.java
```diff
@@ -140,9 +140,15 @@ private IRubyObject consumeHash(int size) {
     RubyHash hash = RubyHash.newHash(runtime);
     for (int i = 0; i < size; i++) {
       IRubyObject key = next();
-      if (this.symbolizeKeys && key instanceof RubyString) {
+      if (key instanceof RubyString) {
+        if (this.symbolizeKeys) {
           key = ((RubyString) key).intern();
+        } else {
+          key.setFrozen(true);
+          key = runtime.freezeAndDedupString((RubyString) key);
```
tagomoris, line 148: Is it needed to do `key.setFrozen(true)` even before `runtime.freezeAndDedupString((RubyString) key)`?

ahorek: looking at the source, I think you're right and it's redundant https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/Ruby.java#L4830
btw I copied it from https://github.com/msgpack/msgpack-ruby/blob/master/ext/java/org/msgpack/jruby/Decoder.java#L125

ahorek: removed via https://github.com/msgpack/msgpack-ruby/pull/229/commits/a607b4496589840be7706f8bf0243cb82ecea829

tagomoris: @ahorek Hey, that commit does not belong to this pull-req.

Issue #239, opened 2022-01-20
Latest msgpack (1.4.3) broken on JRuby when using Java 8 VM


ivoanjo:
Hello! I work on Datadog's [ddtrace gem](https://github.com/DataDog/dd-trace-rb) and noticed this issue [on our CI](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/5167/workflows/1808cc8b-e055-479c-a50d-f00469b260fb/jobs/193173) today.

The latest msgpack version, when run with the Java 8 VM breaks:

```
$ docker run jruby:9.3-jre8 bash -c "gem install msgpack; ruby -rmsgpack -e 'puts RUBY_DESCRIPTION; [1,2,3].to_msgpack'"
/opt/jruby/lib/ruby/stdlib/socket.rb:4: warning: already initialized constant Socket::Constants::AF_INET6
/opt/jruby/lib/ruby/stdlib/socket.rb:4: warning: already initialized constant Socket::Constants::IPPROTO_ICMPV6
/opt/jruby/lib/ruby/stdlib/socket.rb:4: warning: already initialized constant Socket::Constants::IPPROTO_IPV6
/opt/jruby/lib/ruby/stdlib/socket.rb:4: warning: already initialized constant Socket::Constants::PF_INET6
/opt/jruby/lib/ruby/stdlib/socket.rb:4: warning: already initialized constant Socket::Constants::SOL_IPV6
Successfully installed msgpack-1.4.3-java
1 gem installed
jruby 9.3.3.0 (2.6.8) 2022-01-19 b26de1f5c5 OpenJDK 64-Bit Server VM 25.312-b07 on 1.8.0_312-b07 +jit [linux-x86_64]
Unhandled Java exception: java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer;
java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer;
           readRubyString at org/msgpack/jruby/Encoder.java:70
                   encode at org/msgpack/jruby/Encoder.java:81
                    write at org/msgpack/jruby/Packer.java:128
                     call at org/msgpack/jruby/Packer$INVOKER$i$1$0$write.gen:-1
             cacheAndCall at org/jruby/runtime/callsite/CachingCallSite.java:372
                     call at org/jruby/runtime/callsite/CachingCallSite.java:175
[...]

PR #240, opened 2022-01-21
restore java 8 compatibility

ahorek:
a dirty fix for https://github.com/msgpack/msgpack-ruby/issues/239

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+Unreleased
+* JRuby - fixes binary compatibility with Java 8
+
 2022-01-20 version 1.4.3:
 
 * Optimize serialization/deserialization of Symbols
--- a/ext/java/org/msgpack/jruby/Decoder.java
+++ b/ext/java/org/msgpack/jruby/Decoder.java
@@ -2,6 +2,7 @@ package org.msgpack.jruby;
 
 
 import java.math.BigInteger;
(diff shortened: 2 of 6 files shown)


tagomoris:
If I understand correctly, casts can't avoid the problem:
```java
# foo.java
class Main {
    public static void main(String[] args) {
        A obj1 = new B();
        obj1.yay();
        B obj2 = new B();
        ((A) obj2).yay();
    }
}

class A {
    public void yay() {
        System.out.println("A");
    }
}

class B extends A {
    public void yay() {
        System.out.println("B");
    }
}
```

```
$ java foo.java
B
B
```

ahorek:
here's an explanation https://www.morling.dev/blog/bytebuffer-and-the-dreaded-nosuchmethoderror/

a ```--release``` flag could prevent a forgotten cast in the future, but rake-compiler doesn't seem to support it https://github.com/rake-compiler/rake-compiler/issues/200

tagomoris:
Supporting `--release` option on rake-compiler looks promising. I'm writing the patch.

tagomoris:
Created https://github.com/rake-compiler/rake-compiler/pull/201

ahorek:
closing in favor of https://github.com/msgpack/msgpack-ruby/pull/241 thanks @tagomoris !

PR #242, opened 2022-01-22
Drop unused directives from gemspec

olleolleolle:
`test_files` is not used by RubyGems.org.


tagomoris:
Thank you! Good to update my knowledge.

Issue #243, opened 2022-01-27
Allow to register native types as extensions


casperisfine:
Related: https://github.com/msgpack/msgpack-ruby/issues/57
Ref: https://github.com/msgpack/msgpack/issues/206

The message pack spec doesn't have support for big nums, which sometimes cause issues when serializing arbitrary data.

It would be nice if we could register a custom type for `Integer`, that would be called when the number is too big, e.g. something like:

```ruby
require 'msgpack'

factory = MessagePack::Factory.new
factory.register_type(
  0x42,
  Integer,
  packer: ->(int) { int.to_s },
  unpacker: -> (payload) { Integer(payload) }
)

factory.dump(42) # Use the native integer type
factory.dump(2**400) # Use the registered packer
```

tagomoris:
I'm not positive to support this feature because `Integer` objects will be packed into 2 (or more?) different msgpack types. It seems too complicated for users. It will happen if the value is too big for `unsigned long long`. What will happen if we register `Float`?
@casperisfine Do you have any use cases other than `Integer`?

tagomoris:
And an idea comes up: we may be able to solve the problem of too big `Integer` by overriding `Integer#to_msgpack_with_packer` method (defined [here](to_msgpack_with_packer(packer))).
This method will just call `packer#write_int` method. We can call `packer#write_ext` in `rescue RangeError` instead.

Yes, it is a little hacky. But IMO it's better to contain this code in this gem as an opt-in module than introducing many complex code on ext type mechanism only for Integer (if there are no other use cases).

The pseudo code is (I don't try to run it):
```ruby
class TooBigInteger
  def initialize(int)
    @int = int
  end
  def to_s
    @int.to_s
  end
end

factory.regsiter_type(0x42, TooBigInteger, packer: ->(obj){ obj.to_s }, unpacker: ->(payload){ Integer(payload) })
class Integer
  private
  def to_msgpack_with_packer(packer)
    packer.write_int self
    packer
  rescue RangeError
    packer.write_ext TooBigInteger(self)
    packer
  end
end
```

casperisfine:
> It seems too complicated for users

I don't think most users would really be concerned about the serialization format. Usually when you use msgpack extensions it's because you are both producer and consumer, so "complexity" of the format doesn't really enter into account. 

> What will happen if we register Float?

Likely nothing, as Float would always fit? Maybe you'd prefer some more explicit `register_bigint` kind of API?

[...]

PR #244, opened 2022-01-27
Support serializing bigint as extension types

casperisfine:
Fix: https://github.com/msgpack/msgpack-ruby/issues/243

```ruby
factory.register_type(
  0x42,
  Integer,
  packer: ->(int) { int.to_s },
  unpacker: -> (payload) { Integer(payload) }
)

factory.dump(42) # Use the native integer type
factory.dump(2**400) # Use the registered packer
```

### DIFF
--- a/ext/java/org/msgpack/jruby/Encoder.java
+++ b/ext/java/org/msgpack/jruby/Encoder.java
@@ -40,10 +40,11 @@ public class Encoder {
   private final ExtensionRegistry registry;
 
   public boolean hasSymbolExtType;
+  private boolean hasBigintExtType;
 
   private ByteBuffer buffer;
 
-  public Encoder(Ruby runtime, boolean compatibilityMode, ExtensionRegistry registry, boolean hasSymbolExtType) {
+  public Encoder(Ruby runtime, boolean compatibilityMode, ExtensionRegistry registry, boolean hasSymbolExtType, boolean hasBigintExtType) {
     this.runtime = runtime;
     this.buffer = ByteBuffer.allocate(CACHE_LINE_SIZE - ARRAY_HEADER_SIZE);
     this.binaryEncoding = runtime.getEncodingService().getAscii8bitEncoding();
@@ -51,6 +52,7 @@ public class Encoder {
     this.compatibilityMode = compatibilityMode;
     this.registry = registry;
     this.hasSymbolExtType = hasSymbolExtType;
+    this.hasBigintExtType = hasBigintExtType;
   }
 
   public boolean isCompatibilityMode() {
@@ -147,7 +149,10 @@ public class Encoder {
     BigInteger value = object.getBigIntegerValue();
     if (value.compareTo(RubyBignum.LONG_MIN) < 0 || value.compareTo(RubyBignum.LONG_MAX) > 0) {
       if (value.bitLength() > 64 || (value.bitLength() > 63 && value.signum() < 0)) {
-        throw runtime.newArgumentError(String.format("Cannot pack big integer: %s", value));
+        if (hasBigintExtType && tryAppendWithExtTypeLookup(object)) {
+          return;
+        }
+        throw runtime.newRangeError(String.format("Cannot pack big integer: %s", value));
       }
       ensureRemainingCapacity(9);
       buffer.put(value.signum() < 0 ? INT64 : UINT64);
@@ -391,14 +396,6 @@ public class Encoder {
 
   private boolean tryAppendWithExtTypeLookup(IRubyObject object) {
     if (registry != null) {
-      RubyModule lookupClass;
-
-      if (object.getType() == runtime.getSymbol()) {
-        lookupClass = object.getType();
-      } else {
-        lookupClass = object.getSingletonClass();
-      }
-
       IRubyObject[] pair = registry.lookupPackerForObject(object);
       if (pair != null) {
         RubyString bytes = pair[0].callMethod(runtime.getCurrentContext(), "call", object).asString();
--- a/ext/msgpack/factory_class.c
+++ b/ext/msgpack/factory_class.c
@@ -31,6 +31,7 @@ typedef struct msgpack_factory_t msgpack_factory_t;
 struct msgpack_factory_t {
     msgpack_packer_ext_registry_t pkrg;
     msgpack_unpacker_ext_registry_t *ukrg;
+    bool has_bigint_ext_type;
     bool has_symbol_ext_type;
     bool optimized_symbol_ext_type;
     int symbol_ext_type;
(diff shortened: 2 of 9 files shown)


ext/msgpack/packer_ext_registry.h
```diff
@@ -89,20 +89,20 @@ static inline VALUE msgpack_packer_ext_registry_lookup(msgpack_packer_ext_regist
...
 
-      type = msgpack_packer_ext_registry_fetch(pkrg, lookup_class, ext_type_result);
+      if (FL_TEST(lookup_class, FL_SINGLETON)) {
+          type = msgpack_packer_ext_registry_fetch(pkrg, lookup_class, ext_type_result);
 
-      if(type != Qnil) {
-          return type;
+          if(type != Qnil) {
+              return type;
+          }
```
casperisfine, line 105: I'll submit part of this as a separate PR, because unless I'm mistaken, this code cause all extension objects to have their `singleton_class` defined when they are serialized, which is pretty bad for performance.

IMO Ruby should expose a way to check if an object has a singleton class without creating it while doing so. I'll open a ticket upstream as well to add such feature. It already exists as `VALUE rb_singleton_class_get(VALUE obj)` but isn't exposed to extensions, and there's no Ruby side equivalent.

tagomoris: Oh, you are already considering this option! I'd checked #245 at first (because it looked more specific/focused change) and noted the same idea there:
https://github.com/msgpack/msgpack-ruby/pull/245#issuecomment-1023779134

tagomoris:
The motivation about BigInt is acceptable for me.
But it's not clear to me why `Integer` is registered as an ext type but `42` should not be encoded into an ext type binary. Can you explain it clearly to users of msgpack-ruby?

In other words, `2 ** 400` is an Integer, `(2**400).is_a?(Integer)` is true and `(2**400).class` is `Integer`. How can we know whether it should be encoded as `int` or `ext 0x42` (in your sample code)?

casperisfine:
> But it's not clear to me why `Integer` is registered as an ext type but `42` should not be encoded into an ext type binary. Can you explain it clearly to users of msgpack-ruby?

Of course. The rationale is that using the native type when the integer fits in it is:

  - more performant.
  - backward compatible.

IMO extension types only make sense if the data doesn't fit a native type, and I see no benefit into being able to replace native types. 

We could however have it has an option, but again, I'd rather not do it unless someone see a use case for it (I don't).

tagomoris:
@casperisfine Could you rebase this change too?
(Sorry, I'll see other pull-requests tomorrow or later)

casperisfine:
Rebased.

> (Sorry, I'll see other pull-requests tomorrow or later)

No need to apologize, thanks for all the reviews!

ext/java/org/msgpack/jruby/Encoder.java
```diff
@@ -147,7 +149,10 @@ private void appendBignum(RubyBignum object) {
     BigInteger value = object.getBigIntegerValue();
     if (value.compareTo(RubyBignum.LONG_MIN) < 0 || value.compareTo(RubyBignum.LONG_MAX) > 0) {
       if (value.bitLength() > 64 || (value.bitLength() > 63 && value.signum() < 0)) {
-        throw runtime.newArgumentError(String.format("Cannot pack big integer: %s", value));
+        if (hasBigintExtType && tryAppendWithExtTypeLookup(object)) {
+          return;
+        }
+        throw runtime.newRangeError(String.format("Cannot pack big integer: %s", value));
```
casperisfine, line 155: The MRI extension raises `RangeError` on bigint, but the JRuby one `ArgumentError`. I assume it is a bug.

ext/msgpack/factory_class.c
```diff
@@ -209,6 +211,10 @@ static VALUE Factory_register_type(int argc, VALUE* argv, VALUE self)
         }
     }
 
+    if (ext_module == rb_cInteger && RB_TEST(rb_hash_aref(options, ID2SYM(rb_intern("bigint"))))) {
```
casperisfine, line 214: No idea what's the best key to pass, so I went with `bigint: true`, but I'm open to suggestions.

casperisfine (commented):
@tagomoris I updated this PR with your comments from #243

casperisfine:
Option renamed as `oversized_integer_extension: true `

tagomoris (changes requested):
Could you change to ... ?:
* raise ArgumentError if `oversized_integer_extension: true` is specified with non-`Integer` class
* not to require `msgpack/bigint` in default because its serialization format is not any agreed format - it should be required by users explicitly

ext/msgpack/packer.h
```diff
@@ -469,9 +472,28 @@ static inline void msgpack_packer_write_fixnum_value(msgpack_packer_t* pk, VALUE
...
 static inline void msgpack_packer_write_bignum_value(msgpack_packer_t* pk, VALUE v)
 {
+    int leading_zero_bits;
+    size_t required_size = rb_absint_size(v, &leading_zero_bits);
+
     if(RBIGNUM_POSITIVE_P(v)) {
+        if(required_size > 8 && pk->has_bigint_ext_type) {
+            if(msgpack_packer_try_write_with_ext_type_lookup(pk, v)) {
+                return;
+            }
```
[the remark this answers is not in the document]
casperisfine: Yes, I wrote a spec for it, that's how I noticed JRuby was inconsistent.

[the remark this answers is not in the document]
tagomoris: Oh, sorry, I want a code comment at just after L482

tagomoris (approved):
LGTM!

I've created a commit to minimize if-clause for older rubies as small as possible.
WDYT? [7f586b4](https://github.com/msgpack/msgpack-ruby/pull/268/commits/7f586b456157f9f36485bb9d032e6a76e43c949e)

casperisfine:
> WDYT?

It cause two methods to be defined instead of one. Not the end of the world, but seems unnecessary. But if you really don't like `def` inside `if`, it's up to you.

tagomoris:
Using `define_method` with Proc is alternative. Hmm.

casperisfine:
I'd rather not, `define_method` create different ISeq that are slower than regular ones defined with `def`.

casperisfine:
Is this only a stylistic issue? Because this pattern is fairly common.

casperisfine:
> define_method create different ISeq that are slower than regular ones defined with def.

You can search for the `bmethod` term (as in block methods)

tagomoris:
I want to decrease points to confuse myself in the future. `def` in `if` makes me confused when I check code without memory of this change (for code about Ruby 2.6, will be EoL next month!)

casperisfine:
Then maybe some extra commenting could help?

tagomoris:
Could you write the implementation for the current Ruby versions at first (with appropriate code comments)?
Then, the compatibility follows.
I think it should decrease my confusion.

casperisfine:
Of course.

casperisfine:
Done.

tagomoris (approved):
LGTM. Thanks!

casperisfine:
Thanks for the merge!

konsolebox:
Any chance a documentation about this can also be added in README? Its dynamic behavior is not easily apparent and it should promote its use.

PR #245, opened 2022-01-27
Avoid creating singleton classes when looking up extensions

casperisfine:
Extracted from: https://github.com/msgpack/msgpack-ruby/pull/244

`rb_singleton_class(VALUE)` creates the singleton class if it doesn't exist yet, which isn't good for perf. First because it allocates and object, but also because as far as I know it bust method caches.

### DIFF
--- a/ext/java/org/msgpack/jruby/ExtensionRegistry.java
+++ b/ext/java/org/msgpack/jruby/ExtensionRegistry.java
@@ -82,26 +82,21 @@ public class ExtensionRegistry {
      * Objects of type Integer (Fixnum, Bignum), Float, Symbol and frozen
      * String have no singleton class and raise a TypeError when trying to get
      * it.
-     *
-     * Since all but symbols are already filtered out when reaching this code
-     * only symbols are checked here.
      */
-    if (!(object instanceof RubySymbol)) {
-      lookupClass = object.getSingletonClass();
-      pair = fetchEntryByModule(lookupClass);
-      if (pair != null) {
-          return pair;
-      }
-    }
-
-    pair = fetchEntryByModule(object.getType());
+    lookupClass = object.getMetaClass();
+    pair = fetchEntryByModule(lookupClass);
     if (pair != null) {
       return pair;
     }
 
-    if (lookupClass == null) {
-      lookupClass = object.getType(); // only for Symbol
+    RubyModule realClass = object.getType();
+    if (realClass != lookupClass) {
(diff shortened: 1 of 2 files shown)


tagomoris:
I can understand the motivation. But at the same time, I have serious hesitation about merging this change because it is touching the MRI internals.

So, I have an idea to realize this optimization in a long way. How about doing the following steps?:

1. issue a ticket about this problem and your motivation at bugs.ruby-lang.org
2. propose a change to ruby/ruby to add (and expose) a rb_* function to check it without creating singleton classes
3. use the added function in msgpack-ruby if exists

I know this requires time and at least a new version of Ruby. But I think you can understand my hesitation too.

casperisfine:
@tagomoris actually I didn't have time to update the PR just yet, but I think we don't need any of that. `rb_class_of(VALUE)` should return the `singleton_class` if there is one, and the "real" class otherwise, so I think we can just use that.

I'll update the PR in a bit.

casperisfine:
Ok, I refactored the code, I believe it should be acceptable now as we only use fully public and documented APIs.

ext/java/org/msgpack/jruby/ExtensionRegistry.java
```diff
@@ -82,25 +82,23 @@ public IRubyObject lookupUnpackerByTypeId(int typeId) {
...
-     * Since all but symbols are already filtered out when reaching this code
-     * only symbols are checked here.
      */
-    if (!(object instanceof RubySymbol)) {
-      lookupClass = object.getSingletonClass();
-      pair = fetchEntryByModule(lookupClass);
-      if (pair != null) {
-          return pair;
-      }
+    lookupClass = object.getMetaClass();
```
casperisfine, line 86: JRuby has similar APIs:

```java
    /**
     * if exist return the meta-class else return the type of the object.
     *
     */
    @Override
    public final RubyClass getMetaClass() {
        return metaClass;
    }

    // ....
 
       /**
     * @see org.jruby.runtime.builtin.IRubyObject#getType()
     */
    @Override
    public final RubyClass getType() {
        return metaClass.getRealClass();
    }
```

So we can use the same strategy, first `getMetaClass` and later `getType`, then check wether we got the same object or not.

casperisfine:
This one just got a conflict, resolving it now.

casperisfine:
Rebased.

tagomoris:
Looks perfect!

PR #246, opened 2022-01-28
Fix `symbolize_keys: true` to properly create UTF-8 symbols

casperisfine:
When `symbolize_keys: true` was set, we'd already read the string as binary, even if the native type is UTF-8. This was causing non-ASCII keys to be improperly deserialized.

`rb_str_intern` will still notice strings that are ASCII only, and return ASCII symbol for them. So there's no backward compatibility concerns.

This fixes: https://github.com/ruby-i18n/i18n/issues/606

### DIFF
--- a/ext/msgpack/buffer.h
+++ b/ext/msgpack/buffer.h
@@ -494,9 +494,9 @@ static inline VALUE msgpack_buffer_read_top_as_string(msgpack_buffer_t* b, size_
 #endif // HAVE_RB_ENC_INTERNED_STR
 }
 
-static inline VALUE msgpack_buffer_read_top_as_symbol(msgpack_buffer_t* b, size_t length)
+static inline VALUE msgpack_buffer_read_top_as_symbol(msgpack_buffer_t* b, size_t length, bool utf8)
 {
-    return rb_str_intern(msgpack_buffer_read_top_as_string(b, length, true, false));
+    return rb_str_intern(msgpack_buffer_read_top_as_string(b, length, true, utf8));
 }
 
 #endif
--- a/spec/unpacker_spec.rb
+++ b/spec/unpacker_spec.rb
@@ -1,5 +1,3 @@
-# encoding: ascii-8bit
-
 require 'stringio'
 require 'tempfile'
 require 'zlib'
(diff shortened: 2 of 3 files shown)


casperisfine:
@tagomoris apologies for the ping, but do you think you'll have time to review this PR soon? 

It's causing issue for people using `i18n 1.9.1` and `bootsnap`, so I'd like to know if I should find some workaround to prevent more people from hitting this bug.

tagomoris:
@casperisfine Probably I'll be able to find time to review changes early next week

ext/msgpack/buffer.h
```diff
@@ -494,9 +494,9 @@ static inline VALUE msgpack_buffer_read_top_as_string(msgpack_buffer_t* b, size_
 #endif // HAVE_RB_ENC_INTERNED_STR
 }
 
-static inline VALUE msgpack_buffer_read_top_as_symbol(msgpack_buffer_t* b, size_t length)
+static inline VALUE msgpack_buffer_read_top_as_symbol(msgpack_buffer_t* b, size_t length, int utf8)
```
tagomoris, line 497: `msgpack_buffer_read_top_as_string` defines the last argument as `bool`.
Is there any reason to use `int` here?

casperisfine: Nope, just an oversight, I'll fix it right now.

casperisfine: Fixed.

spec/unpacker_spec.rb
```diff
@@ -302,6 +300,19 @@ def flush
...
+    loaded_hash = MessagePack.load(MessagePack.pack(hash), :symbolize_keys => true)
+
+    loaded_hash.keys[0].should == hash.keys[0]
+    loaded_hash.keys[0].encoding.should == hash.keys[0].encoding
+
+    loaded_hash.keys[1].should == hash.keys[1]
+    loaded_hash.keys[1].encoding.should == hash.keys[1].encoding
+
+    MessagePack.unpack(MessagePack.pack(hash), :symbolize_keys => true).should == hash
+  end
```
tagomoris, line 314: This test case is not enough clear to me because the encodings of `:ascii` and `:utf8_é` depend on the behavior of Ruby runtime and be are not described in this test case.
IMO it will be more clear to me if the original encoding of serialized symbols and expected encoding of deserialized symbols are written explicitly.
@casperisfine What do you think?

casperisfine: Makes sense. I'll update right away.

casperisfine: How does it look now?

tagomoris:
👍

PR #249, opened 2022-02-04
Optimize `MessagePack_Buffer_set_options`

casperisfine:
### Context

`msgpack` is quite fast at serializing/deserializing, however instantiating the packer and unpacker is relatively slow. So on smaller payloads it often end up slower than much less optimized serializers. So I'm looking at optimizing this part.

### Profiling

Starting with the following script:

```ruby
FACTORY = MessagePack::Factory.new
1_000_000.times do
  FACTORY.unpacker
end
```

Capture d’écran, le 2022-02-04 à 16 09 07

About 20% of the time is spent in `MessagePack_Buffer_set_options` which is surprising.

After a quick look, we're calling `rb_respond_to` a lot, and it isn't very fast. So in the common case when `io` is `nil`, we can take a shortcut.

After this change it's mostly gone (`0.2%`):

Capture d’écran, le 2022-02-04 à 16 10 38

### DIFF
--- a/ext/msgpack/buffer_class.c
+++ b/ext/msgpack/buffer_class.c
@@ -62,24 +62,22 @@ static VALUE Buffer_alloc(VALUE klass)
 
 static ID get_partial_read_method(VALUE io)
 {
-    if(rb_respond_to(io, s_readpartial)) {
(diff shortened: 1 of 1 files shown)


tagomoris:
LGTM. Thanks!

PR #250, opened 2022-02-04
Packer_initialize: lazily allocate the extension and cache hashes

casperisfine:
Each packer eagerly allocate two hashes, one for it's extension registry another to act as a cache for lookup and it's expensive.

Assuming no extension type is used, we can significantly speedup the packer instantiation by skipping these two allocations.

And in case extension types are used, we can still delay the cache allocation until it is actually needed.

benchmark (`Object.new` is used as a comparison point):

```ruby
require "benchmark/ips"
require "msgpack"

FACTORY = MessagePack::Factory.new
Benchmark.ips do |x|
  x.report("Object.new") { Object.new }
  x.report("FACTORY.packer") { FACTORY.packer }
  x.compare!
end
```

before:
```
          Object.new      8.512M (± 0.7%) i/s -     42.709M in   5.017901s
      FACTORY.packer      1.932M (± 0.8%) i/s -      9.697M in   5.019840s

Comparison:
          Object.new:  8511619.9 i/s
      FACTORY.packer:  1931874.0 i/s - 4.41x  (± 0.00) slower
```

after:
```
          Object.new      8.484M (± 1.7%) i/s -     42.680M in   5.032470s
      FACTORY.packer      3.399M (± 1.7%) i/s -     17.159M in   5.049362s

Comparison:
          Object.new:  8483734.5 i/s
      FACTORY.packer:  3399232.3 i/s - 2.50x  (± 0.00) slower
```

### DIFF
--- a/ext/msgpack/packer_ext_registry.c
+++ b/ext/msgpack/packer_ext_registry.c
@@ -30,8 +30,8 @@ void msgpack_packer_ext_registry_static_destroy()
 
 void msgpack_packer_ext_registry_init(msgpack_packer_ext_registry_t* pkrg)
 {
-    pkrg->hash = rb_hash_new();
-    pkrg->cache = rb_hash_new();
+    pkrg->hash = Qnil;
+    pkrg->cache = Qnil;
 }
 
 void msgpack_packer_ext_registry_mark(msgpack_packer_ext_registry_t* pkrg)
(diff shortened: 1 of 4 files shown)


ext/msgpack/packer_ext_registry.c
```diff
@@ -43,15 +43,21 @@ void msgpack_packer_ext_registry_mark(msgpack_packer_ext_registry_t* pkrg)
...
 }
 
 VALUE msgpack_packer_ext_registry_put(msgpack_packer_ext_registry_t* pkrg,
         VALUE ext_module, int ext_type, VALUE proc, VALUE arg)
 {
-    VALUE e = rb_ary_new3(3, INT2FIX(ext_type), proc, arg);
-    /* clear lookup cache not to miss added type */
-    rb_hash_clear(pkrg->cache);
-    return rb_hash_aset(pkrg->hash, ext_module, e);
+
```
tagomoris, line 53: meaningless blank line.

tagomoris (commented):
Looks good except for just a comment 😉

ext/msgpack/packer_ext_registry.c
casperisfine: Removed.

tagomoris:
👍

PR #252, opened 2022-02-08
Fix all compilation warnings

casperisfine:
I'll detail the fix for each in diff comment.

### DIFF
--- a/ext/msgpack/buffer_class.c
+++ b/ext/msgpack/buffer_class.c
@@ -245,10 +245,11 @@ static VALUE read_until_eof_rescue(VALUE args)
     return Qnil;
 }
 
-static VALUE read_until_eof_error(VALUE args)
+static VALUE read_until_eof_error(VALUE args, VALUE error)
 {
     /* ignore EOFError */
     UNUSED(args);
+    UNUSED(error);
     return Qnil;
 }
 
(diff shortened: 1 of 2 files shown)


ext/msgpack/buffer_class.c
```diff
@@ -245,10 +245,11 @@ static VALUE read_until_eof_rescue(VALUE args)
     return Qnil;
 }
 
-static VALUE read_until_eof_error(VALUE args)
+static VALUE read_until_eof_error(VALUE args, VALUE error)
```
casperisfine, line 248: `rb_rescue2` callback takes the rescued error as second argument.

ext/msgpack/unpacker_class.c
```diff
@@ -451,7 +424,6 @@ void MessagePack_Unpacker_module_init(VALUE mMessagePack)
     rb_define_method(cMessagePack_Unpacker, "skip_nil", Unpacker_skip_nil, 0);
     rb_define_method(cMessagePack_Unpacker, "read_array_header", Unpacker_read_array_header, 0);
     rb_define_method(cMessagePack_Unpacker, "read_map_header", Unpacker_read_map_header, 0);
-    //rb_define_method(cMessagePack_Unpacker, "peek_next_type", Unpacker_peek_next_type, 0);  // TODO
```
casperisfine, line 454: This todo is from 2012, so I think we can just remove it. The compiler was complaining `Unpacker_peek_next_type ` is unused.

PR #253, opened 2022-02-08
Optimize Packer and Unpacker initialize methods

casperisfine:
Particularly through the common `MessagePack.load / dump`.

```ruby
require 'msgpack'
payload = MessagePack.dump([1, "2", 3.4])

1_000_000.times do
  MessagePack.load(payload)
end
```

Capture d’écran, le 2022-02-08 à 13 23 22

First we accept `nil` in place of the option hash, which allows to skip several hash lookup when no option is passed.

Then if we actually pass an option hash, we used static symbols instead of goign through the ID2SYM every time.

Finally for MessagePack.pack, we accept `nil` in place of var args, as to avoid a costly `*rest` forwarding.

### DIFF
--- a/ext/java/org/msgpack/jruby/Unpacker.java
+++ b/ext/java/org/msgpack/jruby/Unpacker.java
@@ -56,30 +56,48 @@ public class Unpacker extends RubyObject {
 
   @JRubyMethod(name = "initialize", optional = 2, visibility = PRIVATE)
   public IRubyObject initialize(ThreadContext ctx, IRubyObject[] args) {
+    Ruby runtime = ctx.runtime;
+
     symbolizeKeys = false;
     allowUnknownExt = false;
     freeze = false;
-    if (args.length > 0) {
-      Ruby runtime = ctx.runtime;
-      if (args[args.length - 1] instanceof RubyHash) {
-        RubyHash options = (RubyHash) args[args.length - 1];
-        IRubyObject sk = options.fastARef(runtime.newSymbol("symbolize_keys"));
-        if (sk != null) {
-          symbolizeKeys = sk.isTrue();
-        }
-        IRubyObject f = options.fastARef(runtime.newSymbol("freeze"));
-        if (f != null) {
-          freeze = f.isTrue();
-        }
-        IRubyObject au = options.fastARef(runtime.newSymbol("allow_unknown_ext"));
-        if (au != null) {
-          allowUnknownExt = au.isTrue();
-        }
-      }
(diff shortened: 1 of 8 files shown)


casperisfine:
Before / After:

Capture d’écran, le 2022-02-08 à 13 33 03

casperisfine:
Ah, I forgot about jruby again. I'll fix it.

casperisfine:
This should be ready for review.

ext/msgpack/packer_class.c
```diff
@@ -68,29 +70,20 @@ VALUE MessagePack_Packer_alloc(VALUE klass)
...
-    } else if(argc == 1) {
-        VALUE v = argv[0];
-        if(rb_type(v) == T_HASH) {
-            options = v;
-        } else {
-            io = v;
-        }
+    if (argc > 2) {
+        rb_raise(rb_eArgError, "wrong number of arguments (%d for 0..2)", argc);
+    }
```
tagomoris, line 78: IIUC, Ruby's keyword arguments (`options` here) should be the last argument, not 1st (`argv[0] for argc >= 1`) or 2nd (`argv[1] for argc >= 2`).
So, could you place `if (argc > 2)` clause at first, then check `argc >= 1` and `argc == 2`?

ext/java/org/msgpack/jruby/Unpacker.java
```diff
@@ -59,27 +59,37 @@ public IRubyObject initialize(ThreadContext ctx, IRubyObject[] args) {
...
-          allowUnknownExt = au.isTrue();
-        }
+
+    RubyHash options = null;
+
+    if (args.length >= 2 && args[1] instanceof RubyHash) {
+      options = (RubyHash) args[1];
+    } else if (args.length >= 1 && args[0] instanceof RubyHash) {
+      options = (RubyHash) args[0];
+    }
```
tagomoris, line 69: Similar to the comment on CRuby

tagomoris (commented):
Don't you instantiate symbols (like `compatibility_mode`, `symbolize_keys`, etc) statically?

casperisfine:
> Don't you instantiate symbols (like compatibility_mode, symbolize_keys, etc) statically?

I don't think that's possible:

```c
../../../../ext/msgpack/unpacker_class.c:36:42: error: statement expression not allowed at file scope
static VALUE sym_symbolize_keys = ID2SYM(rb_intern("symbolize_keys"));
```

That's also how it's done across MRI.

tagomoris:
LGTM

PR #254, opened 2022-02-08
Avoid zeroing msgpack_unpacker_t twice

casperisfine:
It's already zero-ed by `ZALLOC`, no need to call `memset(0, ...)` again.

Before / After

Capture d’écran, le 2022-02-08 à 15 31 12

This gets rid of about 5% of the execution time spent in zeroing a second time.


tagomoris:
👍

PR #256, opened 2022-02-08
Implement lazy instantiation and Copy on Write for unpacker registry

casperisfine:
Every time an `Unpacker` is instantiated we have to copy over a `VALUE[256]` so `1kiB`.
That's quite a lot of memory usage for such short lived objects, and copying isn't
that cheap.

So first we can only allocate that array when the first type is registered.
So in cases where no extension is used we simply don't incur the cost for it.

Then if we actually do need one, we can only copy it when it happens to be mutated.
In most cases, types are registered once during boot, but very rarely at runtime. So
in the vast majority of cases we should be able to just pass a pointer.

Every time we share a registry we increment it's `borrow_count`, every time a holder
is garbage collected, it decreate the counter, and the last one is in charge of freeing the registry.

Finally when registering a type in the registry, if it isn't yet shared, we
directly mutate it.

```ruby
require 'benchmark/ips'
require 'msgpack'

payload = MessagePack.dump([1, "2", 3.4])

Benchmark.ips do |x|
  x.report("load") { MessagePack.load(payload) }
  x.compare!
end
```

master:
```
$ ruby -Ilib bench/load.rb
                load    95.076k i/100ms
                load    931.850k (± 3.3%) i/s -      4.659M in   5.005056s
```

this branch:
```
                load   142.119k i/100ms
                load      1.442M (± 2.1%) i/s -      7.248M in   5.027389s
```

### DIFF
--- a/ext/msgpack/factory_class.c
+++ b/ext/msgpack/factory_class.c
@@ -30,7 +30,7 @@ typedef struct msgpack_factory_t msgpack_factory_t;
 
 struct msgpack_factory_t {
     msgpack_packer_ext_registry_t pkrg;
-    msgpack_unpacker_ext_registry_t ukrg;
+    msgpack_unpacker_ext_registry_t *ukrg;
     bool has_symbol_ext_type;
     bool optimized_symbol_ext_type;
     int symbol_ext_type;
@@ -49,14 +49,14 @@ static void Factory_free(msgpack_factory_t* fc)
         return;
     }
     msgpack_packer_ext_registry_destroy(&fc->pkrg);
-    msgpack_unpacker_ext_registry_destroy(&fc->ukrg);
+    msgpack_unpacker_ext_registry_release(fc->ukrg);
     xfree(fc);
 }
 
 void Factory_mark(msgpack_factory_t* fc)
 {
     msgpack_packer_ext_registry_mark(&fc->pkrg);
-    msgpack_unpacker_ext_registry_mark(&fc->ukrg);
+    msgpack_unpacker_ext_registry_mark(fc->ukrg);
 }
 
 static VALUE Factory_alloc(VALUE klass)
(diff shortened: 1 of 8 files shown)


tagomoris:
I'll review this change, but at first, I want a new test case to see two unpackers are not sharing those registries.

casperisfine:
@tagomoris test added!

ext/msgpack/factory_class.c
```diff
@@ -113,9 +112,8 @@ VALUE MessagePack_Factory_unpacker(int argc, VALUE* argv, VALUE self)
 
     msgpack_unpacker_t* uk;
     Data_Get_Struct(unpacker, msgpack_unpacker_t, uk);
-
-    msgpack_unpacker_ext_registry_destroy(&uk->ext_registry);
-    msgpack_unpacker_ext_registry_dup(&fc->ukrg, &uk->ext_registry);
+    msgpack_unpacker_ext_registry_borrow(fc->ukrg);
+    uk->ext_registry = fc->ukrg;
```
tagomoris, line 116: What do you think if `msgpack_unpacker_ext_registry_borrow` will receive `uk->ext_regsitery` pointer and assign `fc->ukrg` into it? I mean L116 should be done in `msgpack_unpacker_ext_registry_borrow` because it should happen always when borrowing `fc->ukrg`.

ext/msgpack/factory_class.c
```diff
@@ -208,7 +209,7 @@ static VALUE Factory_register_type(int argc, VALUE* argv, VALUE self)
         }
     }
 
-    msgpack_unpacker_ext_registry_put(&fc->ukrg, ext_module, ext_type, unpacker_proc, unpacker_arg);
+    fc->ukrg = msgpack_unpacker_ext_registry_put(fc->ukrg, ext_module, ext_type, unpacker_proc, unpacker_arg);
```
tagomoris, line 212: On this change, `msgpack_unpacker_ext_registry_put` doesn't put an entry into `fc->ukrg` but actually copy and add and entry then return it.
So I think it's better to replace `fc->ukrg` with the newer one directly. Or to rename it.

tagomoris (commented):
Generally LGTM.
But I have some concerns on implementation as commented.

tagomoris:
Great optimization. Thank you!

PR #258, opened 2022-02-09
Exclude time edge cases where the full epoch is not supported.

aardvark179:


spec/timestamp_spec.rb
```diff
@@ -87,14 +87,14 @@
 
     let(:time96_min) { Time.at(-2**63).utc }
     it 'is serialized into timestamp96' do
-      skip if IS_JRUBY # JRuby cannot handle numerics larger than long
+      skip if IS_JRUBY || IS_TRUFFLERUBY # JRuby and TruffleRuby both use underlying Java time classes that do not cover the full 64 bit epoch.
```
eregon, line 90: I don't think it relates directly to epoch seconds.
The limit seems actually:
```java
public final class Year
        implements Temporal, TemporalAdjuster, Comparable<Year>, Serializable {

    /**
     * The minimum supported year, '-999,999,999'.
     */
    public static final int MIN_VALUE = -999_999_999;
    /**
     * The maximum supported year, '+999,999,999'.
     */
    public static final int MAX_VALUE = 999_999_999;
```

So maybe:
JRuby and TruffleRuby both use underlying Java time classes that do not support |year| >= 1 billion

aardvark179: Okay, changed.

eregon:
Could this be merged please? cc @tagomoris

tagomoris:
Oh, sorry, I missed this change.

eregon:
Thanks! Here is a follow-up, which makes the job fully green: https://github.com/msgpack/msgpack-ruby/pull/282

PR #259, opened 2022-02-10
Remove RUBINIUS specific code

casperisfine:
I just noticed it, and I doubt anyone is using this as rubinius hasn't bene updated in a very long time.


tagomoris:
This will be merged at the same time with #213

PR #260, opened 2022-02-10
Do not intern extension type strings

casperisfine:
Fix: https://github.com/msgpack/msgpack-ruby/pull/248#issuecomment-1034704382

Extensions might not expect it, and it's unlikely that they'll be kept as is as binary strings anyway.

So we should use this optimized path only for `RAW_TYPE_STRING` and `RAW_TYPE_BINARY`.


tagomoris:
👍

Issue #224, opened 2021-06-18
Feature Request: Alternative `register_type` API allowing one level of recursion


casperisfine:
Note: I'm saying feature request but I'm willing to implement it myself. I had a quick look and I think it's doable, but I'd rather get a bit of feedback first as it's non-trivial.

### Context

msgpack's extension types are great, but sometimes a bit limited. It work well for simple value types like `Time` etc, but in many cases serializing a complex object to a binary string is tricky.

What we've been doing with @shioyama is basically use msgpack in msgpack, eg.:

```ruby
MyFactory.register_type(
  0x12,
  SomeObject,
  packer: ->(obj) { MyFactory.dump(obj.to_h) },
  unpacker ->(payload) { SomeObject.load(MyFactory.load(payload)) }
)
```

The above is total pseudo code, but hopefully it help conveying the idea.

The problem with that is that it instantiate many new packers/unpackers for individual values, it is also a bit tricky to ensure we don't recurse.

### Feature Request

It would be great if msgpack had a similar feature built-in, as I think it could be implemented much more efficiently. e.g something like:

```ruby
MyFactory.register_meta_type(
  0x12,
  SomeObject,
  packer: &:as_msgpack,
  unpacker: ->(hash) SomeObject.load_from_msgpack(hash),
)
```

The `packer` would return simple Ruby types, and msgpack would take care of serializing them. Of course to avoid unbound recursion, it wouldn't be allowed to return "meta_types", only builtin and regular extension types.

@tagomoris is this a feature you'd be open to if I were to open a PR for it?

tagomoris:
IIUC, this feature request is for a kind of container objects, like Hash and Array.
In the case of Hash and Array, msgpack-ruby calls `#to_msgpack` of contained objects recursively, and `#to_msgpack` receives `Packer` or `IO`(-like) objects as its argument. It can prevent instantiating packers recursively.
(See https://github.com/msgpack/msgpack-ruby/blob/master/lib/msgpack/core_ext.rb )

So, is it possible and useful for the case here?
If `packer` and `unpacker` procs can receive `self` object (`Packer` or `Unpacker` in those context) as `self` or the 2nd argument, we can use it in the recursive packing/unpacking without registering special types or introducing special mechanism for recursive packing/unpacking.

@casperisfine What do you think about this idea?

casperisfine:
[...]

PR #261, opened 2022-02-10
Allow extension procs to operate the packer/unpacker manually

casperisfine:
Proof of concept implementation for https://github.com/msgpack/msgpack-ruby/issues/224#issuecomment-1034518270

The goal here is mostly to prove the implementation is possible and wouldn't require too much code.

I'm not sure wether the API is the best that could be done, if we wish to move the idea forward we should take the time to think wether that "proc arity" based behavior is what we want.

cc @tagomoris

### DIFF
--- a/ext/java/org/msgpack/jruby/Encoder.java
+++ b/ext/java/org/msgpack/jruby/Encoder.java
@@ -38,13 +38,16 @@ public class Encoder {
   private final Encoding utf8Encoding;
   private final boolean compatibilityMode;
   private final ExtensionRegistry registry;
+  private final Packer packer;
 
   public boolean hasSymbolExtType;
   private boolean hasBigintExtType;
+  private boolean recursiveExtension;
 
   private ByteBuffer buffer;
 
-  public Encoder(Ruby runtime, boolean compatibilityMode, ExtensionRegistry registry, boolean hasSymbolExtType, boolean hasBigintExtType) {
+  public Encoder(Ruby runtime, Packer packer, boolean compatibilityMode, ExtensionRegistry registry, boolean hasSymbolExtType, boolean hasBigintExtType) {
+    this.packer = packer;
     this.runtime = runtime;
     this.buffer = ByteBuffer.allocate(CACHE_LINE_SIZE - ARRAY_HEADER_SIZE);
     this.binaryEncoding = runtime.getEncodingService().getAscii8bitEncoding();
@@ -68,9 +71,17 @@ public class Encoder {
   }
 
   private IRubyObject readRubyString() {
-    IRubyObject str = runtime.newString(new ByteList(buffer.array(), 0, buffer.position(), binaryEncoding, false));
-    buffer.clear();
-    return str;
+    if (recursiveExtension) {
+      // If recursiveExtension is true, it means we re-entered encode, so we MUST NOT flush the buffer.
+	  // Instead we return an empty string to act as a null object for the caller. The buffer will actually
+	  // be flushed once we're done serializing the recursive extension.
+	  // All other method that consume the buffer should do so through readRubyString or implement the same logic.
+      return runtime.newString();
+    } else {
+      IRubyObject str = runtime.newString(new ByteList(buffer.array(), 0, buffer.position(), binaryEncoding, false));
+      buffer.clear();
+      return str;
+    }
   }
 
   public IRubyObject encode(IRubyObject object) {
@@ -396,11 +407,30 @@ public class Encoder {
 
   private boolean tryAppendWithExtTypeLookup(IRubyObject object) {
     if (registry != null) {
-      IRubyObject[] pair = registry.lookupPackerForObject(object);
-      if (pair != null) {
-        RubyString bytes = pair[0].callMethod(runtime.getCurrentContext(), "call", object).asString();
-        int type = (int) ((RubyFixnum) pair[1]).getLongValue();
-        appendExt(type, bytes.getByteList());
+      ExtensionRegistry.ExtensionEntry entry = registry.lookupExtensionForObject(object);
+      if (entry != null) {
+        IRubyObject proc = entry.getPackerProc();
+        int type = entry.getTypeId();
+
+        if (entry.isRecursive()) {
+          ByteBuffer oldBuffer = buffer;
+          buffer = ByteBuffer.allocate(CACHE_LINE_SIZE - ARRAY_HEADER_SIZE);
+          recursiveExtension = true;
+
+          ByteList payload;
+          try {
+            IRubyObject args[] = { object, packer };
+            proc.callMethod(runtime.getCurrentContext(), "call", args);
+            payload = new ByteList(buffer.array(), 0, buffer.position(), binaryEncoding, false);
+          } finally {
+            recursiveExtension = false;
+            buffer = oldBuffer;
+          }
+          appendExt(type, payload);
+        } else {
+          RubyString bytes = proc.callMethod(runtime.getCurrentContext(), "call", object).asString();
+          appendExt(type, bytes.getByteList());
+        }
         return true;
       }
     }
--- a/ext/msgpack/packer.c
+++ b/ext/msgpack/packer.c
@@ -121,19 +121,56 @@ void msgpack_packer_write_hash_value(msgpack_packer_t* pk, VALUE v)
 #endif
 }
 
+struct msgpack_call_proc_args_t;
+typedef struct msgpack_call_proc_args_t msgpack_call_proc_args_t;
+struct msgpack_call_proc_args_t {
+    VALUE proc;
+    VALUE arg;
+    VALUE packer;
+};
+
+VALUE msgpack_packer_try_calling_proc(VALUE value)
+{
+    msgpack_call_proc_args_t *args = (msgpack_call_proc_args_t *)value;
+    return rb_funcall(args->proc, s_call, 2, args->arg, args->packer);
+}
+
 bool msgpack_packer_try_write_with_ext_type_lookup(msgpack_packer_t* pk, VALUE v)
 {
-    int ext_type;
+    int ext_type, ext_flags;
 
-    VALUE proc = msgpack_packer_ext_registry_lookup(&pk->ext_registry, v, &ext_type);
+    VALUE proc = msgpack_packer_ext_registry_lookup(&pk->ext_registry, v, &ext_type, &ext_flags);
 
-    if(proc != Qnil) {
+    if(proc == Qnil) {
+        return false;
+    }
+
+    if(ext_flags & MSGPACK_EXT_RECURSIVE) {
+        msgpack_buffer_t parent_buffer = pk->buffer;
+        msgpack_buffer_init(PACKER_BUFFER_(pk));
+
+        int exception_occured = 0;
+        msgpack_call_proc_args_t args = { proc, v, pk->to_msgpack_arg };
+        rb_protect(msgpack_packer_try_calling_proc, (VALUE)&args, &exception_occured);
+
+        if (exception_occured) {
+            msgpack_buffer_destroy(PACKER_BUFFER_(pk));
+            pk->buffer = parent_buffer;
+            rb_jump_tag(exception_occured); // re-raise the exception
+        } else {
+            VALUE payload = msgpack_buffer_all_as_string(PACKER_BUFFER_(pk));
+            StringValue(payload);
+            msgpack_buffer_destroy(PACKER_BUFFER_(pk));
+            pk->buffer = parent_buffer;
+            msgpack_packer_write_ext(pk, ext_type, payload);
+        }
+    } else {
         VALUE payload = rb_funcall(proc, s_call, 1, v);
         StringValue(payload);
         msgpack_packer_write_ext(pk, ext_type, payload);
-        return true;
     }
-    return false;
+
+    return true;
 }
 
 void msgpack_packer_write_other_value(msgpack_packer_t* pk, VALUE v)
--- a/ext/msgpack/unpacker.c
+++ b/ext/msgpack/unpacker.c
@@ -166,14 +166,17 @@ static inline int object_complete_ext(msgpack_unpacker_t* uk, int ext_type, VALU
         return object_complete_symbol(uk, rb_str_intern(str));
     }
 
-    VALUE proc = msgpack_unpacker_ext_registry_lookup(uk->ext_registry, ext_type);
+    int ext_flags;
+    VALUE proc = msgpack_unpacker_ext_registry_lookup(uk->ext_registry, ext_type, &ext_flags);
+
     if(proc != Qnil) {
-        VALUE obj = rb_funcall(proc, s_call, 1, str);
+        VALUE obj;
+        obj = rb_funcall(proc, s_call, 1, str == Qnil ? rb_str_buf_new(0) : str);
         return object_complete(uk, obj);
     }
 
     if(uk->allow_unknown_ext) {
-        VALUE obj = MessagePack_ExtensionValue_new(ext_type, str);
+        VALUE obj = MessagePack_ExtensionValue_new(ext_type, str == Qnil ? rb_str_buf_new(0) : str);
         return object_complete(uk, obj);
     }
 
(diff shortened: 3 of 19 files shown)


casperisfine:
I rebased this. 

If you are ok with the interface, I can implement the JRuby side of it.

tagomoris:
@casperisfine The interface looks good to me as far as I checked the spec (not reviewed the implementation yet - will do after JRuby implementation will arrive).

spec/factory_spec.rb
```diff
@@ -320,6 +320,32 @@ def to_msgpack_ext
...
+      it 'can receive the packer as argument' do
+        factory = MessagePack::Factory.new
+        factory.register_type(0x00, Symbol)
+        factory.register_type(
+          0x01,
+          Point,
+          packer: ->(point, packer) do
+            packer.write(point.to_a)
+          end,
+          unpacker: ->(payload, unpacker) do
```
casperisfine, line 338: I must admit I'm a bit annoyed by this `payload` param being useless.

ext/msgpack/unpacker.c
```diff
@@ -168,12 +170,31 @@ static inline int object_complete_ext(msgpack_unpacker_t* uk, int ext_type, VALU
...
     VALUE proc = msgpack_unpacker_ext_registry_lookup(uk->ext_registry, ext_type);
     if(proc != Qnil) {
-        VALUE obj = rb_funcall(proc, s_call, 1, str);
-        return object_complete(uk, obj);
+        long arity = 0;
+        if (rb_obj_class(proc) == rb_cProc) {
+            arity = rb_proc_arity(proc);
+        } else if (rb_obj_class(proc) == rb_cMethod) {
+            arity = FIX2LONG(rb_funcall(proc, s_arity, 0));
+        }
```
casperisfine, line 178: I really don't have having to check the arity before invoking procs, I think we should check it once when the extension is registered, and store it in the registry.

casperisfine:
I implemented the JRuby version.

ext/msgpack/unpacker.c
casperisfine: > I really don't have having 
I meant don't like

tagomoris:
I updated the title because this change looks not PoC :D

casperisfine:
Ahaha :)

ext/msgpack/unpacker.c
```diff
@@ -168,12 +170,31 @@ static inline int object_complete_ext(msgpack_unpacker_t* uk, int ext_type, VALU
...
+        if (arity == 2) {
+            uk->last_object = Qnil;
+            reset_head_byte(uk);
+            VALUE obj = rb_funcall(proc, s_call, 2, str, uk->buffer.owner);
+
+            if(uk->freeze) {
+                rb_obj_freeze(obj);
+            }
+            uk->last_object = obj;
+            return PRIMITIVE_OBJECT_COMPLETE;
```
tagomoris, line 189: Why don't you call `object_complete(uk, obj)` here instead of executing a part of it?
I'm not sure about this, but It seems better to do `reset_head_byte(uk);` after `rb_funcall(...)` because the state of `uk` should be clean at the end of deserialization under the unpacker Proc's responsibility.

ext/msgpack/packer.h
```diff
@@ -357,6 +357,13 @@ static inline void msgpack_packer_write_map_header(msgpack_packer_t* pk, unsigne
 
 static inline void msgpack_packer_write_ext(msgpack_packer_t* pk, int ext_type, VALUE payload)
 {
+    if (payload == Qnil) {
+        msgpack_buffer_ensure_writable(PACKER_BUFFER_(pk), 3);
+        msgpack_buffer_write_2(PACKER_BUFFER_(pk), 0xc7, 0);
+        msgpack_buffer_write_1(PACKER_BUFFER_(pk), ext_type);
+        return;
+    }
```
tagomoris, line 365: I can't understand why this is added. Could you explain why this is needed?

ext/msgpack/unpacker.c
casperisfine: IIRC it's because I need the `reset_head_byte` to be called, otherwise any write the proc may do is broken. I'll double check though as it's been a few days already.

ext/msgpack/packer.h
casperisfine: It's for this call: https://github.com/msgpack/msgpack-ruby/pull/261/files/727edafda4e2b44e53a2b800dea7b286735c1865#diff-c177757220580d2500819916eee890a84548208c724d85c67a5eb41bde0f8906R135

We pass `Qnil` as a payload to signal that we're in "manual" serialization. When it happens we write a single byte extension that contains a NULL byte.

ext/msgpack/unpacker.c
casperisfine: Yeah I confirm, here's how it fail if I call the proc immediately:

```
  1) MessagePack::Factory#register_type registering ext type with manual serialization can receive the packer as argument (proc)
     Failure/Error: attrs = unpacker.read
     
     EOFError:
       end of buffer reached
     # ./spec/factory_spec.rb:340:in `read'
     # ./spec/factory_spec.rb:340:in `block (5 levels) in <top (required)>'
     # ./lib/msgpack/factory.rb:70:in `full_unpack'
     # ./lib/msgpack/factory.rb:70:in `load'
     # ./spec/factory_spec.rb:347:in `block (4 levels) in <top (required)>'
```

tagomoris: I can understand it should call `reset_head_byte(uk);` before calling the proc.
And I also think that it looks better to call it again after returning from calling the proc. And if it's not required, calling `reset_head_byte(uk);` twice shouldn't be so serious overhead (but just calling `object_complete()` looks very clean).

ext/msgpack/packer.h
tagomoris: Oh, it means, this new feature writes the size `0` for this ext type object, right?
Because it cannot determine its size before calling the Proc.

Wait. It should break other msgpack deserializer implementations...

casperisfine: > Oh, it means, this new feature writes the size 0 for this ext type object, right?

Yes, exactly.

> Wait. It should break other msgpack deserializer implementations...

Well, my reasoning was that if the other implementation don't know about this extension type, they can't deserialize it anyway (unless they have some kind of basic handling of unknown extension types).

ext/msgpack/unpacker.c
casperisfine: You are right, it's much cleaner this way.

ext/msgpack/packer.h
tagomoris: The [spec of type `fext 8`](https://github.com/msgpack/msgpack/blob/master/spec.md#ext-format-family) says:
* ext 8 stores an integer and a byte array whose length is upto (2^8)-1 bytes
* XXXXXXXX is a 8-bit unsigned integer which represents N
* N is a length of data

If an implementation sets `0` into N, other implementations will stop reading bytes at the end of 3rd bytes, then it will try to deserialize the next objects - and all following objects will not be deserialized. It should be helped.

The reason for the heading length for ext types is, unpackers will be able to skip reading those bytes if it doesn't know what the ext type is. And the following bytes (other objects) will be deserialized after skipping it. It is the conversion that all packers/unpackers should follow.

I'm sorry that I missed this point at the time when I foud my original idea. I'll try to find workarounds or any other ideas.

casperisfine: No worries. I thought it was ok. 

Well, let me know if you find another idea. 

Otherwise we might have to buffer what the extension writes to be able to set the proper size, which might be tricky.

tagomoris: My very rough idealist (have never concerned any feasibilities):
* initialize a new buffer (B) in the packer, call the Proc with the packer with B, then restore the original buffer (A), write the length and type-id, concatenate B into A
* create a recorder object (new class, X) which behaves like a Packer, call the Proc with X, then the packer calculates the length to be appended and replay the recorded write operations on the buffer

(The `Recorder` object is like an object, which has methods like `write_nil`, `write_int`, ... and just stores the list of methods and arguments called. It can calculate the byte size to be appended, and also call the series of methods with arguments on the Packer.)

casperisfine: > initialize a new buffer (B) in the packer

That sounds doable.

Another idea would be to make sure the buffer isn't flushed, and then find a way to go back and adjust the size, no idea if it's doable though.

tagomoris: After a quick code check, I found [this line](https://github.com/msgpack/msgpack-ruby/blob/master/ext/msgpack/buffer_class.c#L194). My current preferable idea is:

1. initialize a new Buffer (B) in the packer
2. replace internal buffer (A) with B
3. call the Proc
4. restore A in the packer
5. write the length of B, ext type id
6. append the content of B

We may be able to optimize the Buffer-Buffer append operation as commented there (I don't have solid idea for now).

tagomoris: > Another idea would be to make sure the buffer isn't flushed, and then find a way to go back and adjust the size, no idea if it's doable though.

This requires a completely different usage of Buffer, so I'm not positive about this.

casperisfine: > My current preferable idea is:

Makes sense. I'll try to find some time to implement this.

casperisfine: Ok, I did a quick hack just to see how it would work. The code is disgusting, and probably has a bunch of bugs, but it seems to work for simple cases.

That will definitely need a sizeable refactoring to be clean though :/

tagomoris: I roughly checked the lines of `packer.c` L142-161 and it looks good to me.
Under this concept, these if-clause for `payload == nil` should be removed, right?

casperisfine: > Under this concept, these if-clause for `payload == nil` should be removed, right?

Yes.

casperisfine:
Ok, I managed to update the Java extension as well.

spec/factory_spec.rb
casperisfine: A solution might be to pass an option when registering the extension instead or relying on the `arity`. e.g.

```ruby
        factory.register_type(
          0x01,
          Point,
          packer: ->(point, packer) do
            packer.write(point.to_h)
            nil
          end,
          unpacker: ->(unpacker) do
            attrs = unpacker.read
            Point.new(attrs.fetch(:x), attrs.fetch(:y), attrs.fetch(:z))
          end,
          streaming: true,
        )
```

(doubt `streaming` is a good name but...)

ext/java/org/msgpack/jruby/Encoder.java
```diff
@@ -38,13 +38,16 @@
   private final Encoding utf8Encoding;
   private final boolean compatibilityMode;
   private final ExtensionRegistry registry;
+  private final Packer packer;
 
   public boolean hasSymbolExtType;
   private boolean hasBigintExtType;
+  private boolean streaming;
```
tagomoris, line 45: It's not clear to me what "streaming" means for now. In the serialization/deserialization context, "streaming" means a streaming encoder/decoder like this:
```
decoder = Decoder.new
decoder.register_callback do |decoded_object|
  # ...
end
decoder << str1
decoder << str2
decoder << str3
```

ext/java/org/msgpack/jruby/Encoder.java
```diff
@@ -68,9 +71,13 @@ private void ensureRemainingCapacity(int c) {
...
-    IRubyObject str = runtime.newString(new ByteList(buffer.array(), 0, buffer.position(), binaryEncoding, false));
-    buffer.clear();
-    return str;
+    if (streaming) {
+      return runtime.newString();
+    } else {
+      IRubyObject str = runtime.newString(new ByteList(buffer.array(), 0, buffer.position(), binaryEncoding, false));
+      buffer.clear();
+      return str;
+    }
```
tagomoris, line 80: I want a code-comment here about why it will not read a String when "streaming".

ext/msgpack/packer.c
```diff
@@ -121,19 +121,56 @@ void msgpack_packer_write_hash_value(msgpack_packer_t* pk, VALUE v)
 #endif
 }
 
+struct msgpack_call_proc_args_t;
+typedef struct msgpack_call_proc_args_t msgpack_call_proc_args_t;
```
tagomoris, line 125: Does it have any effect than defining a usual struct `msgpack_call_proc_args_t`?

ext/msgpack/unpacker_class.c
```diff
@@ -162,7 +163,7 @@ static VALUE Unpacker_buffer(VALUE self)
 static VALUE Unpacker_read(VALUE self)
 {
     UNPACKER(self, uk);
-
+
```
tagomoris, line 166: Unexpected space here.

ext/java/org/msgpack/jruby/Encoder.java
casperisfine: I'm not particularly happy with this name either. `streaming` is meant to convey that at this stage the size isn't know yet, the extension callback will "stream" some bytes, that we'll measure once it's done.

If you have idea of a better name, let me know i"ll happily rename it.

ext/msgpack/packer.c
casperisfine: I don't think so no, it's just a common way to pass more than one argument through `rb_protect`, you can see this pattern a bit everywhere in MRI and most extensions.

ext/java/org/msgpack/jruby/Encoder.java
casperisfine: Comment added.

ext/java/org/msgpack/jruby/Encoder.java
tagomoris: How about `swapped` or `temp_buffer`?

casperisfine: Maybe `recursive`?

On a similar note we should probably pass a keyword argument rather than rely on arity: https://github.com/msgpack/msgpack-ruby/pull/261#discussion_r824519919, so we should probably select a user facing name for this concept.

tagomoris: A longer name sounds good to me if we'll use it for a keyword argument :P
`recursive_extension_handling` ... hmm, not beautiful.

casperisfine: `recursive` would look like this:
```ruby
        factory.register_type(
          0x01,
          Point,
          packer: ->(point, packer) { packer.write(point.to_h) },
          unpacker: ->(unpacker) { Point.new(unpacker.read) },
          recursive: true,
        )
```

Since it's an argument to `register_type`, it's implied that it means `recursive_type`.

We could use `recursiveExtensionType` internally?

casperisfine: Or just `recursiveExtension`?

tagomoris: > Since it's an argument to `register_type`, it's implied that it means `recursive_type`.

👍 

> We could use `recursiveExtensionType` internally?

👍

casperisfine: Ok, doing that then.

casperisfine:
Ok, now both the MRI and Java extensions have been updated to directly rely on the `recursive: true` option passed to `Factory#register_type`. The code is much simpler now.

ext/msgpack/unpacker.c
```diff
@@ -40,6 +41,7 @@ void msgpack_unpacker_static_init()
     msgpack_rmem_init(&s_stack_rmem);
 #endif
 
+    s_arity = rb_intern("arity");
```
tagomoris, line 44: ditto

ext/msgpack/unpacker_ext_registry.c
```diff
@@ -18,11 +18,13 @@
 
 #include "unpacker_ext_registry.h"
 
+static ID s_arity;
 static ID s_call;
 static ID s_dup;
 
 void msgpack_unpacker_ext_registry_static_init()
 {
+    s_arity = rb_intern("arity");
```
tagomoris, line 27: ditto

spec/factory_spec.rb
```diff
@@ -320,6 +320,89 @@ def to_msgpack_ext
         expect(factory.dump(42)).to eq(MessagePack.dump(42))
       end
     end
+
+    describe "registering ext type with recursive serialization" do
+      before do
+        stub_const("Point", Struct.new(:x, :y, :z))
+      end
+
```
tagomoris, line 328: This test doesn't check the byte order itself. If both packer&unpacker places bytes in the order of `typeid` + `length`, those tests will pass.
Could you add a test case (or assertions) to check the actual bytes from packers?

tagomoris (commented):
Totally LGTM.
I left some comments about unused code and test cases.

spec/factory_spec.rb
casperisfine: I think the `'sets the correct length'` test handle that, by asserting that if it's parsed as an unknown ext it still works.

But I added an extra test wich directly test the bytes prefix.

ext/java/org/msgpack/jruby/Encoder.java
```diff
@@ -68,9 +71,15 @@ private void ensureRemainingCapacity(int c) {
   }
 
   private IRubyObject readRubyString() {
-    IRubyObject str = runtime.newString(new ByteList(buffer.array(), 0, buffer.position(), binaryEncoding, false));
-    buffer.clear();
-    return str;
+    if (recursiveExtension) {
+      // if recursiveExtension is true, it means we re-entered encode,
+      // so we MUST NOT flush the buffer.
+      return runtime.newString();
```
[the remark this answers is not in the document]
tagomoris: > Because this method is the only place the buffer is consumed. All other reads call this one.

Ah, I see. I was not sure about it. Could you update this comment to add "All other reads should call this one" or something else?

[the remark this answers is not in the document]
casperisfine: Done.

tagomoris (approved):
Great. Thank you!
I'll merge this change and release a new version tomorrow.

casperisfine:
❤️ Thanks for the patience and all the reviews!

PR #262, opened 2022-02-10
No need to call `rb_intern()` during initialization

casperisfine:
That macro has an optimization so that it's properly
cached when called with a constant string:

```
/* __builtin_constant_p and statement expression is available
 * since gcc-2.7.2.3 at least. */
    (RBIMPL_CONSTANT_P(str) ? \
     __extension__ ({ \
         static ID rbimpl_id; \
         rbimpl_intern_const(&rbimpl_id, (str)); \
     }) : \
     (rb_intern)(str))
```

### DIFF
--- a/ext/msgpack/buffer.c
+++ b/ext/msgpack/buffer.c
@@ -19,15 +19,9 @@
 #include "buffer.h"
 #include "rmem.h"
 
-#ifndef HAVE_RB_STR_REPLACE
-static ID s_replace;
-#endif
-
 int msgpack_rb_encindex_utf8;
 int msgpack_rb_encindex_usascii;
 int msgpack_rb_encindex_ascii8bit;
-
-ID s_uminus;
 
 #ifndef DISABLE_RMEM
 static msgpack_rmem_t s_rmem;
@@ -35,7 +29,6 @@ static msgpack_rmem_t s_rmem;
 
 void msgpack_buffer_static_init()
 {
-    s_uminus = rb_intern("-@");
 
     msgpack_rb_encindex_utf8 = rb_utf8_encindex();
     msgpack_rb_encindex_usascii = rb_usascii_encindex();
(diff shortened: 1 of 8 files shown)


ext/msgpack/buffer.c
```diff
@@ -19,23 +19,16 @@
...
 
-ID s_uminus;
-
 #ifndef DISABLE_RMEM
 static msgpack_rmem_t s_rmem;
 #endif
 
 void msgpack_buffer_static_init()
 {
-    s_uminus = rb_intern("-@");
```
tagomoris, line 32: It should be helpful if this whitespace is also deleted.

ext/msgpack/buffer.h
```diff
@@ -485,7 +483,7 @@ static inline VALUE msgpack_buffer_read_top_as_string(msgpack_buffer_t* b, size_
 #endif //STR_UMINUS_DEDUPE_FROZEN
         // MRI 2.5 and older do not deduplicate strings that are already
         // frozen.
-        result = rb_funcall(result, s_uminus, 0);
+        result = rb_funcall(result, rb_intern("-@"), 0);
```
tagomoris, line 486: Does this change mean that unpacking frozen strings will call `#-@` every time?

tagomoris: It looks not a bad idea to create the symbol in the same way with #253
WDYT?

tagomoris (commented):
The lazy-ed `rb_intern()` can be called many times on the workloads of users who use those features.
Can't we help such situations? For example, can we create those symbols statically like #262 ?

ext/msgpack/buffer.h
casperisfine: That's what this PR is about. 

Some functions like `rb_funcall` don't take a Symbol (`VALUE`), they take an `ID` which can be created statically through `rb_intern()`, and Ruby's macro take care of it.

If you look at MRI's source, you'll see that it rarely bother pre-initializing `ID`.

But it's not like it's hurting either, if you like it that way, it's fine I can close.

casperisfine:
So just to be sure I ran a simple benchmark: https://github.com/Shopify/msgpack-ruby/commit/edd17b4b0fef8327a3795f3782326ceccdd2d0da

Not precomputing `rb_intern` is indeed a bit slower. So since it doesn't improve readability that much, I'm not so certain this change is worth it.

PR #264, opened 2022-02-11
Use TypedData structs

casperisfine:
The API is a bit more modern and open the door to several features such as `memsize`, write barriers etc.

The main addition for now is support for `memsize_of`:

```ruby
>> ObjectSpace.memsize_of(f)
=> 2128
```

Note: `Buffer` is still "untyped" because there is actually two types of buffer. Some with their own memory region, and in charge of freeing it, and some that point to the `Unpacker` memory and need to not have a `free` function. 

For now I don't see how to do this with `TypedData` structs.

### DIFF
--- a/ext/msgpack/extconf.rb
+++ b/ext/msgpack/extconf.rb
@@ -56,6 +56,8 @@ rescue NoMethodError
   $CFLAGS << ' -DSTR_UMINUS_DEDUPE_FROZEN=0 '
 end
 
+$CFLAGS << ' -DHAS_GC_COMPACT' if GC.respond_to?(:compact)
+
 if warnflags = CONFIG['warnflags']
   warnflags.slice!(/ -Wdeclaration-after-statement/)
 end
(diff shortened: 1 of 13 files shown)


tagomoris:
Honestly speaking, I have almost no idea about `TypeData` structs.
Could you point any kind of documents or texts we can learn it?

casperisfine:
> Could you point any kind of documents or texts we can learn it?

I'll try to find some more, the MRI C API is unfortunately very undocumented. In the meantime there's this ticket (in japanese so I don't know what it says): https://bugs.ruby-lang.org/issues/1637

casperisfine:
A few more resources:

  - The official extension doc: https://github.com/ruby/ruby/blob/master/doc/extension.rdoc
  - Some unofficial (but generally more helpful) API guide: https://silverhammermba.github.io/emberb/c/#data
  - A somewhat "tutorially" description of it: https://blog.peterzhu.ca/ruby-c-ext-part-7/

chrisseaton:
We should get on and do this - the old data API is very deprecated, and causes problems for future GC.

byroot:
Agreed. I may need to do it one class at a time though because this was too big of a change for me.

I'll try to get back at this this week.

PR #270, opened 2022-03-08
Fix MessagePack::Bigint::TYPE being too large

casperisfine:
I noticed this mistake while integrating it.

I can also remove that constant, it's merely a suggestion so that different app can more easily interoperate.


tagomoris:
Wow, I missed that value.
I want it to be removed, or use a positive int. Minus type ids are reserved by spec, and should not be used for types not written in the spec.

casperisfine:
Ok. I'll remove it.

PR #273, opened 2022-04-06
Fix recursive extensions nesting

casperisfine:
I discovered a bug while integrating https://github.com/msgpack/msgpack-ruby/pull/261

I'm trying to figure out a fix, but so far I must admit I don't quite understand what's happening. It has something to do with the stack being set with `type = STACK_TYPE_ARRAY` which cause the native array to be returned instread of the return value of the extension proc.

I'll keep digging, but figured I should let you know of this bug sooner rather than later. cc @tagomoris

### DIFF
--- a/ext/msgpack/unpacker.c
+++ b/ext/msgpack/unpacker.c
@@ -52,6 +52,16 @@ void msgpack_unpacker_static_destroy()
 
 #define HEAD_BYTE_REQUIRED 0xc1
 
+static inline msgpack_unpacker_stack_t* _msgpack_unpacker_new_stack(void) {
+#ifdef UNPACKER_STACK_RMEM
+    return msgpack_rmem_alloc(&s_stack_rmem);
+    /*memset(uk->stack, 0, MSGPACK_UNPACKER_STACK_CAPACITY);*/
+#else
+    /*uk->stack = calloc(MSGPACK_UNPACKER_STACK_CAPACITY, sizeof(msgpack_unpacker_stack_t));*/
+    return xmalloc(MSGPACK_UNPACKER_STACK_CAPACITY * sizeof(msgpack_unpacker_stack_t));
+#endif
+}
+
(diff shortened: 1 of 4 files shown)


casperisfine:
@tagomoris it's specific to `recursive: true`, other use cases aren't impacted.

tagomoris:
@casperisfine I see. Thanks.

casperisfine:
Ok, I figured the fix, It's a bit quick & dirty for now, but I'll clean it up.

I need to fix the Java version as well that seem to have a similar problem.

tagomoris:
Great. I think it's not needed to yank 1.5.0 because it's only about the newly added feature. 
I'll review and merge&release it tomorrow if it'll be ready.

ext/msgpack/unpacker.c
```diff
@@ -297,7 +304,22 @@ static inline int read_raw_body_begin(msgpack_unpacker_t* uk, int raw_type)
...
             size_t ext_size = uk->reading_raw_remaining;
             uk->reading_raw_remaining = 0;
+
+            msgpack_unpacker_stack_t* stack = uk->stack;
+            size_t stack_depth = uk->stack_depth;
+            size_t stack_capacity = uk->stack_capacity;
+
+            uk->stack = _msgpack_unpacker_new_stack();
+            uk->stack_depth = 0;
+            uk->stack_capacity = MSGPACK_UNPACKER_STACK_CAPACITY;
```
casperisfine, line 314: For the C extension, the problem was different. Before yielding to the extension, we must reset the stack. Or more accurately, set a new fresh stack and restore it later.

casperisfine (commented):
It should be ready for review.

tagomoris (approved):
LGTM. Nice catch!

tagomoris:
I've released 1.5.1. Thank you!

Issue #265, opened 2022-02-11
`Compaction isn't available on this platform`


pvalena:
Hello,

while doing update to msgpack 1.4.4 for Fedora, I've encountered issue on ppc64le (exclusively). Part of Fedora package build, we run the test suite.
I'm hitting similar issue to https://bugs.ruby-lang.org/issues/18133:
```
+ rspec -Ilib -I/builddir/build/BUILDROOT/rubygem-msgpack-1.4.4-1.fc37.ppc64le/usr/lib64/gems/ruby/msgpack-1.4.4 spec
An error occurred while loading ./spec/cases_spec.rb.
Failure/Error: GC.verify_compaction_references(double_heap: true, toward: :empty)
NotImplementedError:
  Compaction isn't available on this platform
# <internal:gc>:257:in `verify_compaction_references'
# ./spec/spec_helper.rb:21:in `<top (required)>'
# ./spec/cases_spec.rb:1:in `<top (required)>'
An error occurred while loading ./spec/cruby/buffer_io_spec.rb.
Failure/Error: GC.verify_compaction_references(double_heap: true, toward: :empty)
NotImplementedError:
  Compaction isn't available on this platform
# <internal:gc>:257:in `verify_compaction_references'
# ./spec/spec_helper.rb:21:in `<top (required)>'
# ./spec/cruby/buffer_io_spec.rb:1:in `<top (required)>'
An error occurred while loading ./spec/cruby/buffer_spec.rb.
Failure/Error: GC.verify_compaction_references(double_heap: true, toward: :empty)
NotImplementedError:
  Compaction isn't available on this platform
# <internal:gc>:257:in `verify_compaction_references'
# ./spec/spec_helper.rb:21:in `<top (required)>'
# ./spec/cruby/buffer_spec.rb:1:in `<top (required)>'
An error occurred while loading ./spec/cruby/unpacker_spec.rb.
Failure/Error: GC.verify_compaction_references(double_heap: true, toward: :empty)
NotImplementedError:
  Compaction isn't available on this platform
# <internal:gc>:257:in `verify_compaction_references'
# ./spec/spec_helper.rb:21:in `<top (required)>'
# ./spec/cruby/unpacker_spec.rb:2:in `<top (required)>'
An error occurred while loading ./spec/ext_value_spec.rb.
Failure/Error: GC.verify_compaction_references(double_heap: true, toward: :empty)
NotImplementedError:
  Compaction isn't available on this platform
[...]

PR #275, opened 2022-04-07
Handle platforms without compaction support

casperisfine:
Fix: https://github.com/msgpack/msgpack-ruby/issues/265

That's a test only concern, so no big deal.

cc @pvalena


pvalena:
Thanks!

chrisseaton (approved):
I think `GC.respond_to?(:verify_compaction_references)` is the better pattern, but not a big deal.

PR #276, opened 2022-04-13
Fix recursive extension parsing skipping too many bytes

casperisfine:
Fix: https://github.com/msgpack/msgpack-ruby/pull/261

The recursive call to `read` already consume the buffer so we shouldn't skip anything.

Apologies for not catching this either.

cc @tagomoris


tagomoris:
Oops, I missed this change.
This looks good... will see #283 and merge&release this then

tagomoris:
I've released v1.5.2 with this fix. Thank you!

PR #281, opened 2022-05-11
Fix `freeze` option to also dedup empty strings

casperisfine:
While profiling our application, I noticed that MessagePack was producing a lot of duplicated empty strings:

```
Retained String Report
-----------------------------------
  73.38 kB    1834  ""
              1593  msgpack-1.5.1/lib/msgpack/factory.rb:167
```

Because empty strings were handled distinctly, it wouldn't go through the deduplication path.

Overall I don't think this optimization is really worth it. Just removing this code is simpler I think.

cc @tagomoris

### DIFF
--- a/ext/msgpack/buffer.h
+++ b/ext/msgpack/buffer.h
@@ -49,6 +49,10 @@
 
 #define NO_MAPPED_STRING ((VALUE)0)
 
+#ifndef RB_ENC_INTERNED_STR_NULL_CHECK
+#define RB_ENC_INTERNED_STR_NULL_CHECK 0
+#endif
+
 extern int msgpack_rb_encindex_utf8;
 extern int msgpack_rb_encindex_usascii;
 extern int msgpack_rb_encindex_ascii8bit;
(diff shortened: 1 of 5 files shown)


casperisfine:
Ok, it was fixed in https://github.com/ruby/ruby/commit/edf01d4e82d8e44ee30ec41fbcb7f802bc8b8c5d, I'll request for it to be backported to 3.0, and add a workaround in the PR.

ext/msgpack/buffer.h
```diff
@@ -456,7 +460,11 @@ static inline VALUE msgpack_buffer_read_top_as_string(msgpack_buffer_t* b, size_
 
 #ifdef HAVE_RB_ENC_INTERNED_STR
     if (will_be_frozen) {
-        result = rb_enc_interned_str(b->read_buffer, length, utf8 ? rb_utf8_encoding() : rb_ascii8bit_encoding());
+        if (RB_ENC_INTERNED_STR_NULL_CHECK && length == 0) {
+            result = rb_enc_interned_str("", length, utf8 ? rb_utf8_encoding() : rb_ascii8bit_encoding());
```
chrisseaton, line 464: If we have just two frozen, interned instances of empty strings in these two encodings should they be store it somewhere and return it rather than running this code each time?

chrisseaton (approved):
Approved, with suggestion.

ext/msgpack/buffer.h
casperisfine: Yeah, would be worth memoizing it in the `init`.

casperisfine: Actually I remember now. This function is inline in a header, I don't know how to have a global variable here. Using `static` wouldn't work.

byroot:
Merging as is for now, but if you have an idea how to precompute these strings let me know and I'll happily throw another PR.

PR #282, opened 2022-05-13
Run rake spec instead of rake and add truffleruby release in the non-head jobs

eregon:
I noticed truffleruby fails in https://github.com/msgpack/msgpack-ruby/runs/6416790321?check_suite_focus=true during doc generation, but doc generation is not relevant for being to run this gem correctly. So `rake spec` is a better target.
The failure is also completely hidden due to ` || echo "failed, but ignore it"`.

Since `truffleruby` it the latest truffleruby and not a head version, and it should be stable, I moved it to the stable releases job.


tagomoris:
@eregon Thanks!

PR #288, opened 2022-07-14
Re-do benchmarks

chrisseaton:
The previous benchmark utility, Viiite, hasn't been updated in a decade. It's not compatible with modern Ruby.

### DIFF
--- a/bench/bench.rb
+++ b/bench/bench.rb
@@ -47,44 +47,32 @@ extended_packer.register_type(0x00, Extended, :to_msgpack_ext)
 data_extended = extended_packer.pack(object_extended).to_s
 
 Benchmark.ips do |x|
-  x.report('pack-plain') do |n|
-    n.times do
-      MessagePack.pack(object_plain)
-    end
+  x.report('pack-plain') do
+    MessagePack.pack(object_plain)
   end
 
-  x.report('pack-structured') do |n|
-    n.times do
-      MessagePack.pack(object_structured)
-    end
+  x.report('pack-structured') do
+    MessagePack.pack(object_structured)
   end
 
-  x.report('pack-extended') do |n|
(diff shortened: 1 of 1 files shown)


byroot:
Agreed.

bench/bench.rb
```diff
@@ -0,0 +1,90 @@
...
+  'extended' => Extended.new
+}
+
+extended_packer = MessagePack::Packer.new
+extended_packer.register_type(0x00, Extended, :to_msgpack_ext)
+data_extended = extended_packer.pack(object_extended).to_s
+
+Benchmark.ips do |x|
+  x.report('pack-plain') do |n|
+    n.times do
```
byroot, line 51: Does using explit `times` makes an accuracy difference? I always assumed that that equivalent to the default iteration method.

byroot (dismissed):
1 question, otherwise LGTM

bench/bench.rb
eregon: Agreed, this should be removed, it just adds more overhead and noise.

byroot: I suppose it can make sense for snipets that are very fast to use a `while (n =- 1) > 0` do save on the block calls etc. But I don't think these benchmarks qualify.

eregon:
> It's not compatible with modern Ruby.

What's the error actually, could you open an issue at https://github.com/blambeau/viiite?  Viiite has more features like charts, etc.
But agreed benchmrk-ips is far more common and more maintained, so if people don't use those extra features benchmark-ips is better.

chrisseaton:
> What's the error

https://github.com/blambeau/viiite/issues/6

eregon (approved):
Looks good to me.

byroot (approved):

PR #290, opened 2022-07-20
Limit by io_buffer_size when reading directly to a string

### LINKED ISSUE #118: Invalid argument @ io_getpartial when loading from big files
The following code triggers the error:

``` ruby
require 'msgpack'
require 'fileutils'
require 'securerandom'

TMP_FILE_PATH = "/tmp/testmsgpack1"
MAX_SIZE = 1_000_000 * 3_000 # 3gb
# note that no error is raised if for example MAX_SIZE = 2mb

def generate_random_string
  string = "".force_encoding("BINARY")
  loop do
    string << (SecureRandom.random_bytes(2048)*100)
    return string if string.bytesize >= MAX_SIZE
  end
end

begin
  # Create an hash with 3 keys, the value of the key "c" is a binary string of 3gb
  print "to_msgpack..."
  string = {"a" => "foo", "b" => 42, "c" => generate_random_string}.to_msgpack
  puts "done"

  # Write the msgpack to disk
  print "write to disk..."
  File.open(TMP_FILE_PATH, "wb") do |f|
    curr = 0
    loop do
      buffer = string[curr..curr+20000-1]
      break if buffer.to_s.bytesize == 0
      curr += f.write(buffer)
    end
  end
  puts "done"

  # Load it...
  print "loading..."
  # the next line will trigger `readpartial': Invalid argument @ io_getpartial - /tmp/testmsgpack1 (Errno::EINVAL) 
  MessagePack.load(File.open(TMP_FILE_PATH))
  puts "done"
ensure
  FileUtils.rm_f(TMP_FILE_PATH)
end
```

chrisseaton:
Otherwise for reads larger than `INTMAX` the `read` system call will fail.

Fixes #118.

### DIFF
--- a/ext/msgpack/buffer.c
+++ b/ext/msgpack/buffer.c
@@ -613,9 +613,11 @@ size_t _msgpack_buffer_feed_from_io(msgpack_buffer_t* b)
 
 size_t _msgpack_buffer_read_from_io_to_string(msgpack_buffer_t* b, VALUE string, size_t length)
 {
+#define MIN(x, y) (((x) < (y)) ? (x) : (y))
+
     if(RSTRING_LEN(string) == 0) {
         /* direct read */
-        VALUE ret = rb_funcall(b->io, b->io_partial_read_method, 2, SIZET2NUM(length), string);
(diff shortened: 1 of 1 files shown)


byroot (approved):
👍

Issue #291, opened 2022-07-20
Seg Fault in #load


larry-reid-shopify:
When `MessagePack::DefaultFactory#load` is configured to pack `Symbol`s and is run on the string below, it produces a seg fault.

Ruby version: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]
msgpack version: msgpack-1.5.3

Steps to reproduce:

```
irb -Ilib
require "msgpack"

result = "\x82\xC7\v\x00_generation\xD9(1207eef075b2c65215b9f3050074aa8c13f920b2\xD8\x00translation_data\x81\xC7\x00\x00\xC0"
codec = MessagePack::DefaultFactory
type_attributes = {
  code: 0,
  packer: Symbol.method_defined?(:name) ? :name.to_proc : :to_s.to_proc,
  unpacker: :to_sym.to_proc,
  optimized_symbols_parsing: true,
}
codec.register_type(
  type_attributes.fetch(:code),
  Symbol,
  type_attributes
)
codec.load(result) # This should give the seg fault.
```

Here's the top of the seg fault dump:

```
/home/spin/.bundle/shopify/gems/msgpack-1.5.3/lib/msgpack/factory.rb:70: [BUG] Segmentation fault at 0x0000000000000008
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0023 p:---- s:0116 e:000115 CFUNC  :full_unpack
c:0022 p:0050 s:0112 e:000111 METHOD /home/spin/.bundle/shopify/gems/msgpack-1.5.3/lib/msgpack/factory.rb:70
c:0021 p:0008 s:0105 e:000104 EVAL   (irb):20 [FINISH]
c:0020 p:---- s:0102 e:000101 CFUNC  :eval
c:0019 p:0020 s:0094 e:000093 METHOD /opt/rubies/ruby-3.1.2/lib/ruby/3.1.0/irb/workspace.rb:119
c:0018 p:0054 s:0086 e:000084 METHOD /opt/rubies/ruby-3.1.2/lib/ruby/3.1.0/irb/context.rb:476
c:0017 p:0161 s:0077 e:000076 BLOCK  /opt/rubies/ruby-3.1.2/lib/ruby/3.1.0/irb.rb:577
[...]

PR #292, opened 2022-07-21
Handle empty symbol in optimized symbol parser

casperisfine:
Fix: https://github.com/msgpack/msgpack-ruby/issues/291

To avoid allocating garbage, the parser return Qnil for `ext` types if the byte array is empty.

So we were missing a nil check.

cc @chrisseaton @larry-reid-shopify


chrisseaton (approved):

larry-reid-shopify (approved):
It's been a long time since I was paid to look at C code, but this looks fine to me.

PR #294, opened 2022-08-19
Fix a GC bug in recursive extensions

casperisfine:
Fix: https://github.com/msgpack/msgpack-ruby/pull/293

When we process a recursive extension, we replace the unpack stack by a fresh one. The problem is that the stack contains VALUE references so if GC triggers, we won't mark these objects and they'll br garbage collected.
    
To solve this `msgpack_unpacker_stack_t` now has a `parent` pointer, and the unpacker mark function recursively scan the parent stacks.

I've split the PR in multiple commits so it's easier to review. The first two are a small refactoring to make the change easier, the last one is the actual fix.

### DIFF
--- a/ext/msgpack/unpacker.c
+++ b/ext/msgpack/unpacker.c
@@ -60,7 +60,7 @@ void msgpack_unpacker_static_destroy()
 #define HEAD_BYTE_REQUIRED 0xc1
 
 static inline msgpack_unpacker_stack_t* _msgpack_unpacker_new_stack(void) {
-    msgpack_unpacker_stack_t *stack = ALLOC(msgpack_unpacker_stack_t);
+    msgpack_unpacker_stack_t *stack = ZALLOC(msgpack_unpacker_stack_t);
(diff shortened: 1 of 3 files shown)


ext/msgpack/unpacker.c
```diff
@@ -60,13 +60,16 @@ void msgpack_unpacker_static_destroy()
 #define HEAD_BYTE_REQUIRED 0xc1
 
 static inline msgpack_unpacker_stack_t* _msgpack_unpacker_new_stack(void) {
+    msgpack_unpacker_stack_t *stack = ZALLOC(msgpack_unpacker_stack_t);
```
casperisfine, line 63: We have to do one more alloc, which is a bit unfortunate. In theory we could alloc the stack header and stack data together, but I'd need to dig into how `msgpack_rmem_alloc` works.

byroot:
cc @peterzhu2118

chrisseaton (approved):

PR #296, opened 2022-08-23
Refactor MessagePack::Factory to use TypedData_Get_Struct

casperisfine:
`Data_Get_Struct` is somewhat deprecated.

This is extracted from https://github.com/msgpack/msgpack-ruby/pull/264 so it's easier to review.


byroot:
Also @peterzhu2118 if you don't mind reviewing this, as you're much more familiar with the TypedData API than I am.

ext/msgpack/factory_class.c
```diff
@@ -54,23 +49,54 @@ static void Factory_free(msgpack_factory_t* fc)
...
+
+static const rb_data_type_t factory_data_type = {
+    .wrap_struct_name = "msgpack:factory",
+    .function = {
+        .dmark = Factory_mark,
+        .dfree = Factory_free,
+        .dsize = Factory_memsize,
+    },
+    .flags = RUBY_TYPED_FREE_IMMEDIATELY
+};
```
chrisseaton, line 79: This initialiser syntax isn't incompatible with C compilers that people are still trying to use is it?

chrisseaton (approved):

ext/msgpack/factory_class.c
casperisfine: That's the syntax used in Ruby itself, so I assume it's fine? e.g. https://github.com/ruby/ruby/blob/46c3a93982d0b81668668a65c10bc3670a488d8a/process.c#L604-L611

peterzhu2118 (approved):
👍 This makes sense to me

PR #297, opened 2022-08-23
Fix #127: Unpacker#read_array_header and Unpacker#read_map_header to return correct size

### LINKED ISSUE #127: Unpacker#read_array_header and Unpacker#read_map_header return a wrong size.
In windows environment with msgpack-1.0.2-x86-mingw32, Unpacker#read_array_header for array16 and Unpacker#read_map_header for map16 return a wrong size.

A code to reproduce is following.
```
require 'msgpack'

p MessagePack::VERSION

unpacker = MessagePack::Unpacker.new()
# feed array 16 including one element.
unpacker.feed("\xdc\x00\x01\x01")
# expect "1" but returns "10001"
puts 'read_array_header : ' + unpacker.read_array_header.to_s(16)
unpacker.read.to_s
# feed map 16 including one element.
unpacker.feed("\xde\x00\x02\x01\x02\x03\x04")
# expect "2" but returns "20002"
puts 'read_map_header : ' + unpacker.read_map_header.to_s(16)
```

In Windows7(64bit) with ruby 2.2.1p85 (2015-02-26 revision 49769) [i386-mingw32], the code returns wrong size as below.
 > "1.0.2"
 > read_array_header : 10001
 > read_map_header : 20002

The result of read_array_header is excepted "1" but it returns "0x010001".
The result of read_map_header is excepted "2" but it returns "0x020002".

I guess there are some type-cast or bit-mask miss.

In Linux(CentOS7) with ruby 2.0.0p598 [x86_64-linux], it works correctly.

> "1.0.2"
> read_array_header : 1
> read_map_header : 2

mohits:
Fixes the problem outlined here https://github.com/msgpack/msgpack-ruby/issues/127 by expanding the returned `be16` value safely into a `uint32_t` before writing it to the memory pointed to by `result_size`.

I have verified that the test spec runs correctly on Ruby 2.6.6 (Windows) and the output for the program in #127 is as below (ignore the version since I changed it to be different from the installed version on my computer):

```
"1.5.7"
read_array_header : 1
read_map_header : 2
```

The installed 1.5.6 returns the erroneous output.

```
"1.5.6"
read_array_header : 10001
read_map_header : 20002
```

### DIFF
--- a/ext/msgpack/unpacker.c
+++ b/ext/msgpack/unpacker.c
@@ -678,7 +678,7 @@ int msgpack_unpacker_read_array_header(msgpack_unpacker_t* uk, uint32_t* result_
     } else if(b == 0xdc) {
         /* array 16 */
         READ_CAST_BLOCK_OR_RETURN_EOF(cb, uk, 2);
-        *result_size = _msgpack_be16(cb->u16);
+        *result_size = _msgpack_be16(cb->u16) & 0x0000FFFF; /* expand value to 32 bits */
 
     } else if(b == 0xdd) {
         /* array 32 */
@@ -706,7 +706,7 @@ int msgpack_unpacker_read_map_header(msgpack_unpacker_t* uk, uint32_t* result_si
     } else if(b == 0xde) {
         /* map 16 */
         READ_CAST_BLOCK_OR_RETURN_EOF(cb, uk, 2);
-        *result_size = _msgpack_be16(cb->u16);
+        *result_size = _msgpack_be16(cb->u16) & 0x0000FFFF; /* expand value to 32 bits */
 
     } else if(b == 0xdf) {
         /* map 32 */


mohits:
There's another way to fix this and I wonder which is better. The `#define` for `_msgpack_be16(x)` in `sysdep.h` can produce an incorrect result in a 32-bit integer. The fix is to apply the mask at that level instead as below. Technically, this is the correct implementation for a _be16 and should be correct everywhere this macro is used.

```c
#    define _msgpack_be16(x) ( \
        ( \
        ((((uint16_t)x) <<  8) ) | \
        ((((uint16_t)x) >>  8) ) ) \
        & 0x0000FFFF \
        )
```

ext/msgpack/unpacker.c
```diff
@@ -678,7 +678,7 @@ int msgpack_unpacker_read_array_header(msgpack_unpacker_t* uk, uint32_t* result_
     } else if(b == 0xdc) {
         /* array 16 */
         READ_CAST_BLOCK_OR_RETURN_EOF(cb, uk, 2);
-        *result_size = _msgpack_be16(cb->u16);
+        *result_size = _msgpack_be16(cb->u16) & 0x0000FFFF; /* expand value to 32 bits */
```
byroot, line 681: Shouldn't that be inside `_msgpack_be16`?

mohits: Hi @byroot - you're right. It should be there and I had assumed that it must have been working fine since it's also should been used in multiple places. It seems not.

I will raise a PR with the change as I noted in:
https://github.com/msgpack/msgpack-ruby/pull/297#issuecomment-1224531953

I didn't want to do it at that time of the night.

mohits:
A better fix is in #298.

PR #299, opened 2022-08-24
Refactor MessagePack::Packer to use TypedData_Get_Struct

casperisfine:
Ref: https://github.com/msgpack/msgpack-ruby/pull/264


casperisfine:
So that's where I'm a bit unsure.

IIRC, when using `TypedData`, you must use write barriers (`RB_OBJ_WRITE` and `RB_OBJ_WRITTEN`), but I only have a fairly limited understanding of this.

The `msg_packer_t` struct has two `VALUE` fields:

```c
struct msgpack_packer_t {
    msgpack_buffer_t buffer;

    bool compatibility_mode;
    bool has_bigint_ext_type;
    bool has_symbol_ext_type;

    ID to_msgpack_method;
    VALUE to_msgpack_arg;

    VALUE buffer_ref;

    /* options */
    bool comaptibility_mode;
    msgpack_packer_ext_registry_t ext_registry;
};
```

So I think we should use `RB_OBJ_WRITE` when we write into them, like here:

```c
static inline void msgpack_packer_set_to_msgpack_method(msgpack_packer_t* pk,
        ID to_msgpack_method, VALUE to_msgpack_arg)
{
    pk->to_msgpack_method = to_msgpack_method;
    pk->to_msgpack_arg = to_msgpack_arg;
}
```

The problem is `msgpack` eagerly extract the struct pointer, and then stop dealing with the wrapper object, so we don't have access to it in most places. So if I'm correct this would require some more refactoring.

@peterzhu2118 is my understanding correct?

casperisfine:
Hum, reading https://bugs.ruby-lang.org/projects/ruby-master/wiki/RGenGC, write barriers are not strictly necessary, they are only useful to allow objects to be promoted to the old gen and not be marked on minor GC runs.

If my understanding is correct, we can merge as is, and then look at implementing it in a followup if necessary.

Normally apps shouldn't have too many packer / unpacker / factory instances, but still would be nice to allow them to be promoted etc.

peterzhu2118:
If you don't have `RUBY_FL_WB_PROTECTED` in the `flags` field of `rb_data_typed_t` then the object is WB unprotected, so you don't need to call the WB functions.

byroot:
Ok, so my understanding was correct. Thank you!

byroot (approved):
Leaning on Peter's review.

PR #301, opened 2022-08-24
Refactor MessagePack::Buffer to use TypedData_Get_Struct

casperisfine:
Fix: https://github.com/msgpack/msgpack-ruby/pull/264

This one is a bit more subtle than the other two classes.

The reason is that the Buffer class can be used in two distinct ways.

If you call `Buffer.new`, and independant `msgpack_buffer_t` is allocated. In this case, the class is responsible for freeing it.

However you can also call `Packer#buffer` or `Unpacker#buffer`. In this case, the `msgpack_buffer_t` is actually inline inside the Packer / Unpacker struct. So the class shouldn't free it.

Previously that was done by conditionally passing the `free` function to `Data_Wrap_Struct`, but that's no longer possible with `TypedData_*`.

So instead `Buffer_free` now check wether the `msgpack_buffer_t` has an `owner` member, and use that to know wether it should free the buffer or not when it is collected.

cc @peterzhu2118

### DIFF
--- a/ext/msgpack/buffer_class.c
+++ b/ext/msgpack/buffer_class.c
@@ -34,13 +34,6 @@ static VALUE sym_write_reference_threshold;
 static VALUE sym_io_buffer_size;
 
 
-#define BUFFER(from, name) \
-    msgpack_buffer_t *name = NULL; \
-    Data_Get_Struct(from, msgpack_buffer_t, name); \
-    if(name == NULL) { \
-        rb_raise(rb_eArgError, "NULL found for " # name " when shouldn't be."); \
-    }
-
 #define CHECK_STRING_TYPE(value) \
     value = rb_check_string_type(value); \
     if( NIL_P(value) ) { \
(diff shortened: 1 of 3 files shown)


ext/msgpack/buffer_class.c
```diff
@@ -53,16 +46,40 @@ static void Buffer_free(void* data)
...
     msgpack_buffer_destroy(b);
     xfree(b);
 }
 
+const rb_data_type_t buffer_data_type = {
+    .wrap_struct_name = "msgpack:buffer",
+    .function = {
+        .dmark = msgpack_buffer_mark,
+        .dfree = Buffer_free,
+        .dsize = NULL, // TODO: memsie func would be nice
```
casperisfine, line 63: I need to look at how to measure this properly. I'd prefer to do it in a followup though, it's not a regression since the old API wouldn't provide it either.

peterzhu2118 (approved):
This makes sense to me.

byroot (approved):

PR #302, opened 2022-08-24
Remove some compilation flags

casperisfine:
I was looking into measuring the memory footprint of buffers, and these flags makes the code much harder to understand.

They're not documented in the README, and it's particularly hard to set compilation flags when using bundler & such.

I think it's preferable to reduce the number of compilation conditionals.

@chrisseaton any thoughts?


chrisseaton:
Yes simpler is better.

chrisseaton (approved):

PR #303, opened 2022-09-02
Properly report buffers memory footprint through ObjectSpace.memsize_of

casperisfine:
When the buffer is embeded in a packer or unpacker, we report its size on the owner object.

### DIFF
--- a/ext/msgpack/buffer.c
+++ b/ext/msgpack/buffer.c
@@ -98,6 +98,20 @@ void msgpack_buffer_destroy(msgpack_buffer_t* b)
     }
 }
 
+size_t msgpack_buffer_memsize(const msgpack_buffer_t* b)
+{
+    size_t memsize = 0;
+    msgpack_buffer_chunk_t* c = b->head;
+
+    while(c) {
+        memsize += sizeof(msgpack_buffer_chunk_t);
+        memsize += (c->last - c->first);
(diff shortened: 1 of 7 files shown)


ext/msgpack/buffer.c
```diff
@@ -98,6 +98,20 @@ void msgpack_buffer_destroy(msgpack_buffer_t* b)
...
 }
 
+size_t msgpack_buffer_memsize(const msgpack_buffer_t* b)
+{
+    size_t memsize = 0;
+    msgpack_buffer_chunk_t* c = b->head;
+
+    while(c) {
+        memsize += sizeof(msgpack_buffer_chunk_t);
+        memsize += sizeof(char) * (c->last - c->first);
```
chrisseaton, line 108: `sizeof(char)` is 1 by definition, as `sizeof` returns the size in terms of number of `chars`.

chrisseaton (approved):
You can remove the `sizeof(char)`.

ext/msgpack/buffer.c
byroot: Indeed, I added it for clarity, but if you think it just adds confusion I'm happy to remove it.

byroot:
Done.

chrisseaton (approved):

PR #304, opened 2022-09-13
Restore referenced buffers feature

casperisfine:
Restore referenced buffers feature
    
I think this feature was accidentally removed in https://github.com/msgpack/msgpack-ruby/pull/213
    
On the MRI versions we support `rb_str_dup` won't copy the string but return a shared string. So it's safe to always append a reference when we are provided a string.
    
So instead of copying the RString content into the buffer, we simply keep a reference to it, and point to its RSTRING_PTR().

From some quick benchmarking, it doesn't seem to make a very big difference, but since we already have all the facilities for it, might as well use it.

FYI @chrisseaton (for when you get back, no rush)

### DIFF
--- a/spec/cruby/buffer_spec.rb
+++ b/spec/cruby/buffer_spec.rb
@@ -579,9 +579,11 @@ describe Buffer do
     buffer = MessagePack::Buffer.new
     empty_size = ObjectSpace.memsize_of(buffer)
     expect(empty_size).to be < 400
-    buffer << "a" * 10_000
+    10.times do
+      buffer << "a" * 500
+    end
     memsize = ObjectSpace.memsize_of(buffer)
-    expect(memsize).to be > 10_000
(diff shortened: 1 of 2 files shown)


chrisseaton (dismissed):

byroot (approved):

spec/cruby/buffer_spec.rb
```diff
@@ -579,9 +579,11 @@
     buffer = MessagePack::Buffer.new
     empty_size = ObjectSpace.memsize_of(buffer)
     expect(empty_size).to be < 400
-    buffer << "a" * 10_000
+    10.times do
+      buffer << "a" * 500
+    end
     memsize = ObjectSpace.memsize_of(buffer)
-    expect(memsize).to be > 10_000
```
byroot, line 584: I had to change the test since the string is no longer copied.

PR #308, opened 2022-10-19
Add ruby_memcheck

peterzhu2118:
ruby_memcheck is a tool that helps find memory leaks and errors using Valgrind memcheck. ruby_memcheck helped discover the bug fixed in #307.

This PR adds a `spec:valgrind` target to run the test suite using ruby_memcheck and adds a CI pipeline that runs the target.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,6 +25,17 @@ jobs:
           bundler-cache: true # 'bundle install' and cache
       - run: bundle exec rake
 
+  valgrind:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.1
+          bundler-cache: true # 'bundle install' and cache
+      - run: sudo apt-get install -y valgrind
+      - run: bundle exec rake spec:valgrind
+
   other:
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -1,6 +1,8 @@
 require "set"
 require "objspace"
 
+at_exit { GC.start }
```
byroot, line 4: Is this necessary?

byroot: To clarify, if that's needed, shouldn't `ruby_memcheck` provide a `ruby_memcheck/spec_helper` or something like that to do it automatically?

peterzhu2118: Yes this is needed to run a GC at shutdown to clean up any leftover objects to reduce false positives. It's not in a file that is `require`'d because I don't anticipate ever needing to do anything other than this `GC.start`. ruby_memcheck is designed to be as transparent as possible, so I don't plan on extending or modifying RSpec/minitest in any way.

byroot: Or even better, since you have control over the command: https://github.com/Shopify/ruby_memcheck/blob/a665f906ea5e49fbf4baffd2ae1d6364e924691e/lib/ruby_memcheck/rspec/rake_task.rb#L42 you could pass a `RUBYOPT` to have this code loaded automatically.

byroot (approved):

byroot:
Thank you!

PR #310, opened 2022-12-20
Add a spec to ensure behavior on multiple packers remain stable

casperisfine:
Currently if multiple packers are registered for the same class, the last one is the one that will be used for serialization.

This test is added to make sure this won't change in the future.


casperisfine:
This should go green now, the JRuby error was legit, the way it was implemented would allow for multiple ext_id for the same type.

@peterzhu2118 this should be ready for review.

peterzhu2118 (approved):

PR #317, opened 2023-03-01
Hide all the symbols by defaults to avoid conflict

casperisfine:
We recently experienced segfaults because `buffer_data_type` was name clashing with another gem (https://github.com/msgpack/msgpack-ruby/pull/314#issuecomment-1450366135).

Generally speaking, there is very few use cases for gems to expose any of their symbols, so we should just hide them all by default and avoid this problem entirely.

@peterzhu2118 

CC @flavorjones, as this may be of interest to you.

### DIFF
--- a/ext/msgpack/buffer_class.c
+++ b/ext/msgpack/buffer_class.c
@@ -56,7 +56,7 @@ static size_t Buffer_memsize(const void *data)
     return sizeof(msgpack_buffer_t) + msgpack_buffer_memsize(data);
 }
 
-const rb_data_type_t buffer_data_type = {
+static const rb_data_type_t buffer_data_type = {
     .wrap_struct_name = "msgpack:buffer",
     .function = {
         .dmark = msgpack_buffer_mark,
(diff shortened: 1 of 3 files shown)


peterzhu2118 (approved):

ext/msgpack/buffer_class.c
```diff
@@ -56,7 +56,7 @@ static size_t Buffer_memsize(const void *data)
     return sizeof(msgpack_buffer_t) + msgpack_buffer_memsize(data);
 }
 
-const rb_data_type_t buffer_data_type = {
+static const rb_data_type_t buffer_data_type = {
```
eregon, line 59: I wonder, is the static needed given the -fvisibility=hidden below?

byroot: Yes, but `-fvisibility=hidden` may not work on some platforms, or so I've heard. I still need to cleanup extconf to use `append_cflags` instead of `$CFLAGS <<`

PR #318, opened 2023-03-01
Refactor buffer views

casperisfine:
Two commits for two changes here.

First we no longer set a mark function for `msgpack:buffer_view` objects, instead we mark the `msgpack_buffer_t` struct from the owner (packer or unpacker) mark function. This IMO is much cleaner, and allow us to no longer create the `Buffer` instance eagerly for GC purposes. Given that accessing the buffer is a very rare use case, it is a bit wasteful.

The second change is simply to remove `#dup` and `#clone` on all the classes that never intended to have them defined. Closing buffer, packer or unpacker would copy the content of the `struct`, but that's a shallow copy so it might lead to leaks or use after frees etc.

### DIFF
--- a/lib/msgpack/buffer.rb
+++ b/lib/msgpack/buffer.rb
@@ -1,0 +1,9 @@
+module MessagePack
+  class Buffer
+    # see ext for other methods
+
+    # The semantic of duping a buffer is just too weird.
+    undef_method :dup
+    undef_method :clone
+  end
+end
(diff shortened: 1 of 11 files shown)


lib/msgpack/buffer.rb
```diff
@@ -0,0 +1,32 @@
...
+      when Class
+        klass = klass_or_type
+        registered_types.any?{|entry| klass == entry[:class] }
+      when Integer
+        type = klass_or_type
+        registered_types.any?{|entry| type == entry[:type] }
+      else
+        raise ArgumentError, "class or type id"
+      end
+    end
```
peterzhu2118, line 30: Why are these method added? They don't seem to exist previously and they aren't tested.

byroot: Because I copy pasted like an idiot, and forgot to remove them.

peterzhu2118 (approved):

PR #320, opened 2023-03-13
Unpacker: always feed string references

casperisfine:
Unless you are feeding the unpacker with lots of very small chunks, there is really nothing to gain in copying these strings in the buffer.

We're much better to just hold a CoW string reference that we'll free very soon.

cc @peterzhu2118


ext/msgpack/buffer.c
```diff
@@ -330,14 +330,12 @@ static inline void _msgpack_buffer_append_reference(msgpack_buffer_t* b, VALUE s
...
 void _msgpack_buffer_append_long_string(msgpack_buffer_t* b, VALUE string)
 {
-    size_t length = RSTRING_LEN(string);
-
     if(b->io != Qnil) {
         msgpack_buffer_flush(b);
         if (ENCODING_GET(string) == msgpack_rb_encindex_ascii8bit) {
             rb_funcall(b->io, b->io_write_all_method, 1, string);
         } else {
-            msgpack_buffer_append(b, RSTRING_PTR(string), length);
```
peterzhu2118, line 340: Would it be possible here that `string` gets optimized out and not exist on the stack anymore? Then it could get collected by the GC.

casperisfine: Hum, doubt so, as we'd ultimately always receive it from Ruby code, so it would be referenced there.

peterzhu2118 (approved):

ext/msgpack/buffer.c
casperisfine: Also to clarify, `msgpack_buffer_append(b, RSTRING_PTR(string), RSTRING_LEN(string));` does copy the content.

That's in the case the buffer is linked to an IO, which is relatively rare.

PR #321, opened 2023-03-17
Use sizeof instead of handcrafted arithmetic

casperisfine:
Also removing the `DISABLE_UNPACKER_STACK_RMEM` flag, as I don't see its value. 

@peterzhu2118

### DIFF
--- a/ext/msgpack/unpacker.h
+++ b/ext/msgpack/unpacker.h
@@ -49,7 +49,7 @@ struct msgpack_unpacker_stack_t {
     msgpack_unpacker_stack_t *parent;
 };
 
-#define MSGPACK_UNPACKER_STACK_SIZE (8+4+8+8)  /* assumes size_t <= 64bit, enum <= 32bit, VALUE <= 64bit */
+#define MSGPACK_UNPACKER_STACK_SIZE sizeof(msgpack_unpacker_stack_entry_t)
 
 struct msgpack_unpacker_t {
     msgpack_buffer_t buffer;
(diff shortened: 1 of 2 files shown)


ext/msgpack/unpacker.h
```diff
@@ -49,7 +49,7 @@ struct msgpack_unpacker_stack_t {
     msgpack_unpacker_stack_t *parent;
 };
 
-#define MSGPACK_UNPACKER_STACK_SIZE (8+4+8+8)  /* assumes size_t <= 64bit, enum <= 32bit, VALUE <= 64bit */
+#define MSGPACK_UNPACKER_STACK_SIZE sizeof(msgpack_unpacker_stack_entry_t)
```
peterzhu2118, line 52: This is a good catch, due to alignment `sizeof(msgpack_unpacker_stack_entry_t)` should be 32 rather than 28.

peterzhu2118 (approved):

casperisfine:
Ah damnit, I see why it doesn't use sizeof now, it's because it need that value in the preprocessor.

I still think we can likely clean that up somehow.

peterzhu2118:
This looks good.

> so maybe we'd me better to add a static assertion on msgpack_unpacker_stack_entry_t size

tbh I think that this is a good idea if we never want to turn off `UNPACKER_STACK_RMEM`.

casperisfine:
Closing in favor of https://github.com/msgpack/msgpack-ruby/pull/324

Issue #327, opened 2023-03-27
[BUG] Failed to free an rmem pointer, memory leak?


casperisfine:
We've been digging into this issue with @peterzhu2118 for a couple weeks. We've noticed some crashes in production, but couldn't figure it out.

So we implemented https://github.com/msgpack/msgpack-ruby/pull/323 to make it easier to debug, and now we caught an instance of this bug on CI:

```
[BUG] Failed to free an rmem pointer, memory leak?

-- C level backtrace information -------------------------------------------
/usr/local/ruby/bin/ruby(rb_print_backtrace+0x11) [0x556452739d98] vm_dump.c:785
/usr/local/ruby/bin/ruby(rb_vm_bugreport) vm_dump.c:1080
/usr/local/ruby/bin/ruby(bug_report_end+0x0) [0x5564528d5d0f] error.c:790
/usr/local/ruby/bin/ruby(rb_bug_without_die) error.c:790
/usr/local/ruby/bin/ruby(die+0x0) [0x556452526d39] error.c:798
/usr/local/ruby/bin/ruby(rb_bug) error.c:800
/tmp/bundle/ruby/3.2.0/bundler/gems/msgpack-ruby-51da8d82cb7a/lib/msgpack/msgpack.so(_msgpack_unpacker_destroy+0x6f) [0x7f06b824791f] unpacker.c:78
/tmp/bundle/ruby/3.2.0/bundler/gems/msgpack-ruby-51da8d82cb7a/lib/msgpack/msgpack.so(Unpacker_free+0x21) [0x7f06b8248461] unpacker_class.c:47
/usr/local/ruby/bin/ruby(obj_free+0xec) [0x5564525442ec] gc.c:3606
/usr/local/ruby/bin/ruby(gc_sweep_plane+0x51) [0x55645254bc19] gc.c:5596
/usr/local/ruby/bin/ruby(gc_sweep_page) gc.c:5681
/usr/local/ruby/bin/ruby(gc_sweep_step) gc.c:5992
/usr/local/ruby/bin/ruby(gc_sweep_rest+0x1e) [0x556452553a16] gc.c:6058
/usr/local/ruby/bin/ruby(gc_rest) gc.c:9586
/usr/local/ruby/bin/ruby(gc_rest+0x20) [0x556452556ce7] gc.c:9576
/usr/local/ruby/bin/ruby(objspace_malloc_increase_body) gc.c:12167
[...]

Issue #309, opened 2022-10-26
Packer can overwrite earlier parts of its buffer in version 1.6.0


qcn:
Since upgrading to msgpack 1.6.0 we've been seeing occasional data corruption when packing, which did not happen with 1.5.6. The behaviour is non-deterministic and we haven't been able to explicitly reproduce it yet, but we have several examples of the corrupted data. The symptom is that some bytes from later in the packed data are copied over bytes from earlier, so those bytes appear twice in the packed data (and there is data loss from the corruption).

Example, using Ruby 2.7.6 and msgpack 1.6.0:

The original Ruby object:

```ruby
{"propertyValues"=>
  {"fill"=>nil,
   "stroke"=>"rgb(0,0,0)",
   "fontColor"=>"rgb(0,0,0)",
   "strokeWidth"=>3,
   "stileArrowHead"=>"arrowHeadNone",
   "fontSize"=>16,
   "fontFamily"=>"Open Sans",
   "shapeName"=>"ellipse"},
 "artboard"=>{"width"=>960, "height"=>540},
 "fabricState"=>
  {"objects"=>
    [{"type"=>"upload",
      "originX"=>"center",
      "originY"=>"center",
      "left"=>479.32,
      "top"=>270.38,
      "width"=>1001,
      "height"=>563,
      "fill"=>"rgb(0,0,0)",
      "stroke"=>nil,
      "strokeWidth"=>1,
      "strokeDashArray"=>nil,
      "strokeLineCap"=>"butt",
      "strokeLineJoin"=>"miter",
      "strokeMiterLimit"=>10,
      "scaleX"=>0.96,
      "scaleY"=>0.96,
      "angle"=>0,
      "flipX"=>false,
      "flipY"=>false,
      "opacity"=>1,
      "shadow"=>nil,
      "visible"=>true,
      "clipTo"=>nil,
      "backgroundColor"=>"",
      "fillRule"=>"nonzero",
      "globalCompositeOperation"=>"source-over",
      "stileLocked"=>true,
      "stileUuid"=>"12aad370-b069-47fa-a8ab-a834b8f5903e",
      "stileImageNeedsToBeAutosized"=>false,
      "stileTransientObject"=>false,
      "perPixelTargetFind"=>false,
      "strokeScaledX"=>1,
      "strokeScaledY"=>1,
      "uploadId"=>
       "8224e788d3c957e0749ba7782ae66be5dd68c182d73d60209b2ad49f17b1c705"},
[...]

PR #328, opened 2023-03-27
_msgpack_rmem_alloc2: reset `head.pages` pointer before allocating them.

casperisfine:
`xmalloc` may trigger GC, if it does, some buffers may be freed by ruby and if they were in the previous head, they'll free themselves out of the head.

Fix: https://github.com/msgpack/msgpack-ruby/issues/327

I also suspect this will solve https://github.com/msgpack/msgpack-ruby/issues/309, but I'm not 100% certain (FYI @qcn). 

@peterzhu2118

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+* Fix a possible double-free issue when GC triggers inside `_msgpack_rmem_alloc2`.
 * `Unpacker#feed` now always directly read in provided strings instead of copying content in its buffer.
 * `Unpacker#feed` is now an alias of `Unpacker#feed_reference`.
 * Implement `Factory::Pool#unpacker` and `Factory::Pool#packer` to allow for more precise serialization.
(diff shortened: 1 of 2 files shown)


casperisfine:
As for why this started happening in 1.6.0 and not before:

`1.6.0` is when we migrated to the `TypedData` API, and we did set the `RUBY_TYPED_FREE_IMMEDIATELY` flag. Prior to this the `rmem` pointers wouldn't be freed as part of marking, but later during sweeping. So I think this edge case couldn't happen.

qcn:
Thanks for the tag! We'll upgrade from 1.5.6 when this releases and update on whether the issue appears to be fixed - we were seeing a couple of corruptions a day on average so it should be pretty clear if the issue is still there.

byroot:
We're currently running the master branch in production to make sure the bug we noticed is gone. So far so good.

I may release an 1.7.0 in a couple days, but feel free to try to the master branch until then, your input would be very valuable to confirm the release is sane.

qcn:
Thanks! We'll try master.

qcn:
We've been running ef2f24b660707dc65f329f18242d50300e8e6bf2 for about a month now and haven't seen any re-occurrences of https://github.com/msgpack/msgpack-ruby/issues/309 🎉

PR #332, opened 2023-04-18
Fix an infinite recursion issue when registering a Symbol type with a `nil` packer

casperisfine:
That would cause `has_symbol_ext_type = true` to be set, so `msgpack_packer_write_symbol_value` would invoke `Symbol#to_msgpack` which itself would call back to `write_symbol` causing infinite recursion.

cc @peterzhu2118 @soyed

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,5 @@
+* Fix an infinite recursion issue when registering a Symbol type with a `nil` packer.
+
 2023-03-29 1.7.0:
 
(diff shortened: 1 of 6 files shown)


peterzhu2118 (approved):

soyed (dismissed):

casperisfine:
Urk, of course the Java side need fixing too. I'll deal with it tomorrow.

byroot (approved):

PR #336, opened 2023-05-19
Packer_write_extension: handle T_BIGNUM

casperisfine:
https://github.com/msgpack/msgpack-ruby/pull/335 revealed this issue.

The original code assumed the `ext_type` is a T_FIXNUM.
    
But if the user submit a number large enough, it can be a T_BIGNUM which leads to an assertion error in `FIX2INT()`.


byroot (approved):
FYI: @peterzhu2118

PR #337, opened 2023-05-24
Implement Write Barrier for MessagePack::Factory

casperisfine:
All the references are in `msgpack_unpacker_ext_registry_t` and `msgpack_packer_ext_registry_t` which are shared with `Packer` and `Unpacker` classes.

So we sometimes fire write barriers on these two other types that aren't protected.

### DIFF
--- a/ext/msgpack/packer_ext_registry.c
+++ b/ext/msgpack/packer_ext_registry.c
@@ -28,10 +28,10 @@ void msgpack_packer_ext_registry_static_init(void)
 void msgpack_packer_ext_registry_static_destroy(void)
 { }
 
-void msgpack_packer_ext_registry_init(msgpack_packer_ext_registry_t* pkrg)
+void msgpack_packer_ext_registry_init(VALUE owner, msgpack_packer_ext_registry_t* pkrg)
 {
-    pkrg->hash = Qnil;
-    pkrg->cache = Qnil;
+    RB_OBJ_WRITE(owner, &pkrg->hash, Qnil);
+    RB_OBJ_WRITE(owner, &pkrg->cache, Qnil);
 }
 
 void msgpack_packer_ext_registry_mark(msgpack_packer_ext_registry_t* pkrg)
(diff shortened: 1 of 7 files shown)


ext/msgpack/packer_ext_registry.c
```diff
@@ -40,24 +40,24 @@ void msgpack_packer_ext_registry_mark(msgpack_packer_ext_registry_t* pkrg)
...
-        dst->hash = rb_hash_dup(src->hash);
-        dst->cache = RTEST(src->cache) ? rb_hash_dup(src->cache) : Qnil;
+        RB_OBJ_WRITE(owner, &dst->hash, rb_hash_dup(src->hash));
+        RB_OBJ_WRITE(owner, &dst->cache, rb_hash_dup(src->cache));
     } else {
         // If the type registry is frozen we can safely share it, and share the cache as well.
-        dst->hash = src->hash;
-        dst->cache = src->cache;
+        RB_OBJ_WRITE(owner, &dst->hash, src->hash);
+        RB_OBJ_WRITE(owner, &dst->cache, src->cache);
```
peterzhu2118, line 52: Is this cache always safe to share in this case now? There are now cases where `!RTEST(src->hash) && RTEST(src->cache)` (when the object has just been created in `msgpack_packer_ext_registry_init`).

casperisfine: So if `!RTEST(src->hash)` is true, the `cache` just isn't used at all. But let see if we can make this better by setting it to `Qnil` in that case.

casperisfine: I changed the code to be more explicit.

peterzhu2118: Right, but if we dup a newly created `MessagePack::Factory`, then it will not have a `hash` but have a `cache`. The `hash` is lazily created in the first call to `MessagePack::Factory#register_type` (in `msgpack_packer_ext_registry_put`), which means that two `MessagePack::Factory` could share a `cache` but have completely different `hash`.

ext/msgpack/packer_ext_registry.c
```diff
@@ -40,24 +40,29 @@ void msgpack_packer_ext_registry_mark(msgpack_packer_ext_registry_t* pkrg)
...
+        } else {
+            RB_OBJ_WRITE(owner, &dst->hash, rb_hash_dup(src->hash));
+            RB_OBJ_WRITE(owner, &dst->cache, rb_hash_dup(src->cache));
+        }
     } else {
-        // If the type registry is frozen we can safely share it, and share the cache as well.
-        dst->hash = src->hash;
-        dst->cache = src->cache;
+        RB_OBJ_WRITE(owner, &dst->hash, Qnil);
+        RB_OBJ_WRITE(owner, &dst->cache, Qnil);
```
peterzhu2118, line 57: It looks like some parts of the code (e.g. `msgpack_packer_ext_registry_lookup`) now expect `cache` to always not be `Qnil`. Is it ok to set it to `Qnil` or should we set it to an empty hash?

ext/msgpack/packer_ext_registry.c
casperisfine: Ah I see. Yeah, the issue is that this function is used for `Factory#dup` in which case it should dup both `hash` and `cache` given that it should return a mutable copy of the factory.

I only has the `Factory#packer` callsite in mind. I think it will be simpler to split the two use cases.

ext/msgpack/packer_ext_registry.c
casperisfine: Fixed.

it's unfortunate, but yes, the place where we previously lazily initialized it doesn't have a reference to the owner, so we now have to eagerly create it.

I think if `hash` is `nil` we could forgo creating it though.

PR #338, opened 2023-05-26
Use rb_proc_call* rather than rb_funcall

casperisfine:
It's a tiny bit faster, however it requires the receiver to be an actual Proc.

By refactoring the register_type interface we can eagerly cast all callbacks to actual procs.

In the end it makes the C side of the code simpler which is nice.

There is an extremely minor backward incompatibility in `#registered_types` which no longer return the original callable, but always returns a Proc.

### DIFF
--- a/ext/msgpack/packer_ext_registry.c
+++ b/ext/msgpack/packer_ext_registry.c
@@ -18,16 +18,6 @@
 
 #include "packer_ext_registry.h"
 
-static ID s_call;
-
-void msgpack_packer_ext_registry_static_init(void)
-{
-    s_call = rb_intern("call");
-}
-
-void msgpack_packer_ext_registry_static_destroy(void)
-{ }
-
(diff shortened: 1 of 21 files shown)


ext/msgpack/packer_ext_registry.c
```diff
@@ -18,11 +18,8 @@
 
 #include "packer_ext_registry.h"
 
-static ID s_call;
-
 void msgpack_packer_ext_registry_static_init(void)
 {
-    s_call = rb_intern("call");
 }
```
peterzhu2118, line 23: This is now an empty function, can we remove it?

peterzhu2118 (approved):

ext/msgpack/packer_ext_registry.c
byroot: 👍

peterzhu2118 (approved):

Issue #341, opened 2023-06-28
Attempt to GC mark already marked object


pavel-workato:
Hello! 
When calling MsgPack 1.7.0 we occasionally see following crashes in production. ~ 1 crash on every 1 million calls.

The underlying ruby code is 
```ruby
msgpack = ::MessagePack::Factory.new
msgpack.register_type( ... ) # some custom types for times, symbol, HWIA etc, quite similar to what paquito gem does
msgpack.dump({ ... }) # a pretty big ruby obj, with about 300k keys ~ about 20Mb binary output.
```

Before die VM reports this "[BUG] try to mark T_NONE object". 
```
/app/vendor/bundle/ruby/2.7.0/gems/msgpack-1.7.0/lib/msgpack/factory.rb:77: [BUG] try to mark T_NONE object
ruby 2.7.7p221 (2022-11-24 revision 168ec2b1e5) [x86_64-linux]
-- Control frame information -----------------------------------------------
c:0217 p:---- s:1588 e:001587 CFUNC  :full_pack
c:0216 p:0020 s:1584 e:001583 METHOD /app/vendor/bundle/ruby/2.7.0/gems/msgpack-1.7.0/lib/msgpack/factory.rb:77
...
-- C level backtrace information -------------------------------------------
/app/vendor/ruby-2.7.7/bin/../lib/libruby.so.2.7(rb_print_backtrace+0x15) [0x7fe112f5f615] vm_dump.c:755
/app/vendor/ruby-2.7.7/bin/../lib/libruby.so.2.7(rb_vm_bugreport+0x71b) [0x7fe112f5fd4b] vm_dump.c:941
/app/vendor/ruby-2.7.7/bin/../lib/libruby.so.2.7(rb_bug+0xe4) [0x7fe112d6b778] error.c:645
/app/vendor/ruby-2.7.7/bin/../lib/libruby.so.2.7(gc_mark_maybe+0x0) [0x7fe112dfba90] gc.c:5188
/app/vendor/ruby-2.7.7/bin/../lib/libruby.so.2.7(gc_mark_ptr) (null):0
/app/vendor/bundle/ruby/2.7.0/gems/msgpack-1.7.0/lib/msgpack/msgpack.so(msgpack_buffer_mark+0x21) [0x7fe10b640fc1] buffer.c:119
[...]

PR #344, opened 2023-07-12
Reproducer for issue 341

pavel-workato:
Reproducer for https://github.com/msgpack/msgpack-ruby/issues/341

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,3 +7,4 @@ gemspec
 
 gem "rubocop", "~> 0.82.0"
 gem "simplecov"
+gem "activesupport", "5.2.8.1"
--- a/ext/msgpack/buffer.c
+++ b/ext/msgpack/buffer.c
@@ -113,9 +113,28 @@ size_t msgpack_buffer_memsize(const msgpack_buffer_t* b)
 void msgpack_buffer_mark(void *ptr)
 {
     msgpack_buffer_t* b = ptr;
+    printf("%p: msgpack_buffer_mark\n", b);
     /* head is always available */
     msgpack_buffer_chunk_t* c = b->head;
     while(c != &b->tail) {
+
+        if (RB_TYPE_P(c->mapped_string, T_NONE)) {
+          int i = 0;
+          msgpack_buffer_chunk_t* t = b->head;
+          printf("================ MEGA BUG ==================\n");
+          while(t != &b->tail) {
+            int x = !rb_special_const_p(t->mapped_string) && RB_TYPE_P(t->mapped_string, T_NONE) ? 1 : 0;
+            printf("> %d: %p - %p - %lx - %x\n", i++, b, t, t->mapped_string, (int)(x));
+            t = t->next;
+          }
+          if (b->head != &b->tail) {
+            msgpack_buffer_chunk_t* t = &b->tail;
+            int x = !rb_special_const_p(t->mapped_string) && RB_TYPE_P(t->mapped_string, T_NONE) ? 1 : 0;
+            printf("> %d: %p - %p - %lx - %x\n", i++, b, t, t->mapped_string, (int)(x));
+          }
+          printf("================ END ==================\n");
+        }
+
         rb_gc_mark(c->mapped_string);
         c = c->next;
     }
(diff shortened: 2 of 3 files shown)


casperisfine:
👀

casperisfine:
It crashes on my machine, I'll dig deeper, thank you so much!

casperisfine:
Ok, I've pinpointed the issue, and I got a dirty fix for it. I'll work on a cleaner fix.

The bug is in `msgpack_packer_try_write_with_ext_type_lookup`.

```c
    if(ext_flags & MSGPACK_EXT_RECURSIVE) {
        msgpack_buffer_t parent_buffer = pk->buffer;
        msgpack_buffer_init(PACKER_BUFFER_(pk));

        int exception_occured = 0;
        msgpack_call_proc_args_t args = { proc, { v, pk->to_msgpack_arg } };
        rb_protect(msgpack_packer_try_calling_proc, (VALUE)&args, &exception_occured);

        if (exception_occured) {
            msgpack_buffer_destroy(PACKER_BUFFER_(pk));
            pk->buffer = parent_buffer;
            rb_jump_tag(exception_occured); // re-raise the exception
        } else {
            VALUE payload = msgpack_buffer_all_as_string(PACKER_BUFFER_(pk));
            StringValue(payload);
            msgpack_buffer_destroy(PACKER_BUFFER_(pk));
            pk->buffer = parent_buffer;
            msgpack_packer_write_ext(pk, ext_type, payload);
        }
```

When we call a recursive packer proc, we need to switch the buffer before, and restore it after.

We store it in a local variable, but while it's there, if GC triggers, it's not being marked, so the strings it points to may be GC and the slot set to T_NONE or re-used.

So we need to keep a proper reference to that parent buffer, and mark it.

The bug somewhat exist since the introduction of the `recursive` feature, except when I introduced it `mapped_string` was no longer used, so it would never be reached, but later on I resurected the `mapped_string` feature, opening the door to this bug.

Conceptually the fix is rather simple, but in practice it's made a bit tricker by the fact that:

  - `msgpack_buffer_t` is embeded in the `msgpack_packer_t`.
  - `tail` is embeded inside `msgpack_buffer_t`.

So we can't just store a reference to the copy and make sure `buffer_mark` is called on it. When copy the buffer elsewhere, we need to update some pointers before it's properly markable.

Anyway, I'll work on a clean fix.

sitano:
@casperisfine +1, we also just found that same place with:

```C
            _msgpack_buffer_inspect(&parent_buffer, ">>>>>> before msgpack_buffer");
            VALUE payload = msgpack_buffer_all_as_string(PACKER_BUFFER_(pk)); // <-- GC clears parent_buffer

            StringValue(payload);
            msgpack_buffer_destroy(PACKER_BUFFER_(pk));
            _msgpack_buffer_inspect(&parent_buffer, ">>>>>> before = parent"); // logs T_NONE
```

and the output

```
0x7f31dac870e0: _msgpack_buffer_append_reference: rb_str_dup: 7f31d18fb600
0x7f31dac870e0: msgpack_buffer_all_as_string: start
0x7f31df06f950: >>>>>> before = parent
T_NONE DETECTED 2: 0x7f31df06f950 - 0x7f31dac756a0 - 7f31d18fb600 - 1
```

casperisfine:
Hum, something doesn't quite add up though.

Leaving an unmarked buffer on the stack like this only happens with the `packer`, and I don't get how the packer would have chunks with `mapped_string`.  Unless I'm missing something, it's an `unpacker` only thing.

So I'd like to understand where these `mapped_string` come from 🤔

pavel-workato:
> So I'd like to understand where these mapped_string come from 🤔

From my understanding those strings are already serialised binaries from nested packers. 
Here are a few steps from `msgpack_packer_try_write_with_ext_type_lookup` to the only function assuaging string values for `mapped_string`.

https://github.com/msgpack/msgpack-ruby/blob/96b21a491660f34a7aae16949beff2ca34fcbf9a/ext/msgpack/packer.c#L129-L133

https://github.com/msgpack/msgpack-ruby/blob/96b21a491660f34a7aae16949beff2ca34fcbf9a/ext/msgpack/packer.h#L395

https://github.com/msgpack/msgpack-ruby/blob/96b21a491660f34a7aae16949beff2ca34fcbf9a/ext/msgpack/buffer.h#L258-L261

https://github.com/msgpack/msgpack-ruby/blob/96b21a491660f34a7aae16949beff2ca34fcbf9a/ext/msgpack/buffer.c#L338-L340

https://github.com/msgpack/msgpack-ruby/blob/96b21a491660f34a7aae16949beff2ca34fcbf9a/ext/msgpack/buffer.c#L306-L318

casperisfine:
Yep, I quickly realized I was wrong. When we pack strings, if they are big enough they are added as `mapped_strings`, which saves us from copying them.

https://github.com/msgpack/msgpack-ruby/pull/346 should fix the problem (I don't like the fix, but short of severely refactoring the buffer implementation, that's the most reasonable fix).

PR #347, opened 2023-07-13
Packer: store the parent buffer mapped strings on the stack

casperisfine:
Fix: https://github.com/msgpack/msgpack-ruby/pull/346
Fix: https://github.com/msgpack/msgpack-ruby/pull/344
Fix: https://github.com/msgpack/msgpack-ruby/issues/341

Since the parent buffer isn't marked while we're executing the extension proc, we need to ensure all the chunks `mapped_strings` won't be GCed nor moved.
    
To do that, we copy all of them onto the stack.
    
It's a bit disgusting, but until we refactor the buffer implementation to support recursive extensions better, it's the simplest solution.

cc @peterzhu2118 @pavel-workato @sitano


casperisfine:
Ok, so now MRI CI passes consistently, but interestingly, TruffleRruby still segfault. I'll poke the Truffle team.

pavel-workato:
Thank you! It doesn't crash locally anymore. I'll have a look for possibility to test the fix in production quicker.

ext/msgpack/packer.c
```diff
@@ -114,6 +114,40 @@ bool msgpack_packer_try_write_with_ext_type_lookup(msgpack_packer_t* pk, VALUE v
...
+                mapped_strings_count++;
+            }
+            c = c->next;
+        }
+        if (c->mapped_string != NO_MAPPED_STRING) {
+            mapped_strings_count++;
+        }
+
+        if (mapped_strings_count > 0) {
+            mapped_strings = ALLOCA_N(VALUE, mapped_strings_count);
```
peterzhu2118, line 135: How large can `mapped_strings_count` get? I'm worried about potential stack overflows.

ext/msgpack/packer.c
```diff
@@ -132,6 +166,10 @@ bool msgpack_packer_try_write_with_ext_type_lookup(msgpack_packer_t* pk, VALUE v
             pk->buffer = parent_buffer;
             msgpack_packer_write_ext(pk, ext_type, payload);
         }
+
+        if (mapped_strings_count > 0) {
+            RB_GC_GUARD(mapped_strings[0]);
+        }
```
peterzhu2118, line 172: I'm not sure what the purpose of this `RB_GC_GUARD` is. AFAIK the compiler isn't allowed to reuse the memory allocated in `alloca`.

ext/msgpack/packer.c
byroot: Unfortunately, there isn't really a specific limit.

We could store them in an Array, but then they wouldn't be pinned.

ext/msgpack/packer.c
byroot: Just me having no idea. I'll remove it.

ext/msgpack/packer.c
casperisfine: Yeah, I think you are right.

peterzhu2118 (approved):

ext/msgpack/packer.c
casperisfine: So it turns out clang does optimize the alloca away :/. I'll add this back for now.

casperisfine:
We'll merge this as is for now, but I'm planning to find a better solution soon, probably early next week.

byroot (approved):

eregon (commented):
I think the segfault in https://github.com/msgpack/msgpack-ruby/actions/runs/5540306600/jobs/10112226255 on TruffleRuby is most likely caused by this change.

`ALLOCA_N()` doesn't do any marking on TruffleRuby. And on CRuby alloca() is generally recommended against (can stack overflow, make GC slower, etc).

Could you use a Ruby array instead (`ary = rb_ary_new_capa()`) + `RB_GC_GUARD(ary)`?
That would properly keep those strings alive.

ext/msgpack/packer.c
```diff
@@ -114,6 +114,40 @@ bool msgpack_packer_try_write_with_ext_type_lookup(msgpack_packer_t* pk, VALUE v
     }
 
     if(ext_flags & MSGPACK_EXT_RECURSIVE) {
+        // HACK: While we call the proc, the current pk->buffer won't be reachable
+        // as it will be stored on the stack.
+        // To ensure all the `mapped_string` reference in that buffer are properly
+        // marked and pined, we copy them all on the stack.
```
eregon, line 120: pined -> pinned
BTW I don't see any pinning here, so AFAIK both the `VALUE` strings and their `char*` can still move.

peterzhu2118: All objects on the stack are pinned in Ruby, so all `VALUE` in the alloca will be pinned.

Whereas if we store them in an array, then the pointers can move (since array supports compaction). Only the array will be pinned in this case.

peterzhu2118:
@eregon How does TruffleRuby handle values on the stack? Does it not mark objects that exist on the stack?

eregon:
No, because TruffleRuby (specifically, the JVM GCs) do not scan the native stack or registers (in general/automatically).

Instead it relies on correct usage of RB_GC_GUARD(), etc, and on preserving objects that were converted to a native handle during the current C method call, etc: https://github.com/oracle/truffleruby/blob/master/doc/contributor/cext-values.md

casperisfine:
> Could you use a Ruby array instead (ary = rb_ary_new_capa()) + RB_GC_GUARD(ary)?

As peter said, we can't. But I'm planning to do something similar, just using a custom `RTypedData` that is essentially a pinned Array.

I'm just a bit swamped today, but I do want to do it this week. 

Thanks for having a look at that segfault.

ext/msgpack/packer.c
```diff
@@ -114,6 +114,40 @@ bool msgpack_packer_try_write_with_ext_type_lookup(msgpack_packer_t* pk, VALUE v
     }
 
     if(ext_flags & MSGPACK_EXT_RECURSIVE) {
+        // HACK: While we call the proc, the current pk->buffer won't be reachable
+        // as it will be stored on the stack.
```
eregon, line 118: What do you mean by not reachable?
Maybe this could work through the mark function of `pk`/`pk->buffer`? And ensuring whatever value wraps the `msgpack_packer_t*` is kept alive long enough.

ext/msgpack/packer.c
peterzhu2118: It does not get pinned forever. The pin status of all objects is reset at the start of the GC cycle and gets pinned during marking. Objects on the stack are not treated any differently than other pinned objects.

eregon:
@casperisfine 
> As peter said, we can't.

Why can't we? Because it wouldn't pin and it's needed to pin? (why?)
I'm confused because the comment talks about a Proc and that doesn't access `char*` directly.

Maybe we could use some C API function to pin those mapped strings as soon as they become mapped strings if pinning is required?

ext/msgpack/packer.c
peterzhu2118: `msgpack_buffer_init` resets `pk->buffer`. However, we store the original state of `pk->buffer` on the stack as `parent_buffer`. Since this does not have a corresponding Ruby object, the original values in `pk->buffer` will not get marked.

eregon: Thanks for the explanation.

> Since this does not have a corresponding Ruby object, the original values in `pk->buffer` will not get marked.

Could we create a corresponding Ruby object for it? That seems like a nice solution to get it marked properly.

casperisfine:
https://github.com/msgpack/msgpack-ruby/pull/349 should solve the issue with Truffle (and also is much cleaner in general).

@eregon if you feel like reviewing.

PR #349, opened 2023-07-17
Better fix for marking buffers held on the stack

casperisfine:
Followup: https://github.com/msgpack/msgpack-ruby/pull/347

As discussed previously, the `alloca` hack was just a quick fix, it's not good as it could lead to stack overflows and also could be optimized away by compilers etc.

Here instead we copy all the VALUE references into a dedicated object that is in change of holding and pinning these references.

cc @peterzhu2118 @eregon @pavel-workato @sitano

### DIFF
--- a/ext/msgpack/buffer_class.c
+++ b/ext/msgpack/buffer_class.c
@@ -21,7 +21,8 @@
 #include "buffer.h"
 #include "buffer_class.h"
 
-VALUE cMessagePack_Buffer;
+VALUE cMessagePack_Buffer = Qnil;
+VALUE cMessagePack_HeldBuffer = Qnil;
 
 static ID s_read;
 static ID s_readpartial;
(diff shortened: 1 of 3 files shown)


ext/msgpack/buffer_class.c
```diff
@@ -34,6 +35,73 @@ static VALUE sym_read_reference_threshold;
...
+        return Qnil;
+    }
+
+    msgpack_held_buffer_t* held_buffer = xmalloc(sizeof(msgpack_held_buffer_t) + mapped_strings_count * sizeof(VALUE));
+
+    c = buffer->head;
+    mapped_strings_count = 0;
+    while (c != &buffer->tail) {
+        if (c->mapped_string != NO_MAPPED_STRING) {
+            held_buffer->mapped_strings[held_buffer->size] = c->mapped_string;
```
peterzhu2118, line 92: I think you want to write at `mapped_strings_count`.

```suggestion
            held_buffer->mapped_strings[mapped_strings_count] = c->mapped_string;
```

ext/msgpack/buffer_class.c
```diff
@@ -34,6 +35,73 @@ static VALUE sym_read_reference_threshold;
...
+            mapped_strings_count++;
+        }
+        c = c->next;
+    }
+    if (c->mapped_string != NO_MAPPED_STRING) {
+        held_buffer->mapped_strings[held_buffer->size] = c->mapped_string;
```
peterzhu2118, line 98: Ditto here.

ext/msgpack/buffer_class.c
casperisfine: Yes. I shouldn't write code during meetings 😅 .

peterzhu2118 (approved):

byroot:
> But probably that doesn't quite work and hence the new class?

Yeah that doesn't. The `buffer` struct does this weird thing that it's a linked list of `buffer_chunk_t`, but the `tail` of the list is a member of the `butter_t`. So if you copy the buffer elsewhere, the tail reference is broken...

From what I could gather, it's an optimization to avoid pointer chasing when you append to the buffer, but it's unclear how efficient that really is.

We talked with @peterzhu2118 about potentially simplyfing the buffer implementation, but not sure when we'll actually get around it.

PR #357, opened 2023-12-12
Update README.md example for extension types

ggmichaelgo:
Hello 👋 

I have found that extension type example code in the `README.md` is not working...

Since the `MessagePack#pack` only gives the value to `unpack` and `pack` function:
https://github.com/msgpack/msgpack-ruby/blob/e52d18bacc5a69bd0c823e8daf2792b187a6be46/lib/msgpack.rb#L39-L44

I have updated the example to be functional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,13 +197,11 @@ factory = MessagePack::Factory.new
 factory.register_type(
   0x01,
   Point,
-  packer: ->(point, packer) {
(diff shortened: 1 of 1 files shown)


byroot:
Not sure what you are doing but these examples work fine. Maybe upgrade the gem

PR #371, opened 2024-11-08
Prevent memory leak when a recursive unpacker raises an exception

casperisfine:
The child stack wouldn't be popped nor freed.

Additionally, even once the packer was freed, only the latest stack would be freed, all the parent stack would be leaked.

Practically speaking, every time a recursive unpacker would raise, 4kiB would be leaked.


casperisfine:
I need to check if perhaps we have a similar issue with recursive packers.

PR #370, opened 2024-11-08
Use batch APIs to create arrays and hashes

casperisfine:
Basically a port of https://github.com/ruby/json/pull/678

Rather than to allocate the container and push elements one by one, we accumulate them on a stack and then use the faster batch APIs to directly create the final container.

The original action stack remains, as we need to keep track of what we're currently parsing and how big it is.
    
But for the recursive cases, we no longer need to create a child stack.

PR #372, opened 2024-11-11
Refactor unpacking of recursive types

byroot:
Extracted from: https://github.com/msgpack/msgpack-ruby/pull/370

As initially implemented, we allocate a new stack whenever we encounter a recursive type. This means acquiring a whole 4kiB chunk from the arena, which generally is overkill.

Instead we can introduce a new type of stack item, that behave just like a Hash or Array.

### DIFF
--- a/ext/msgpack/unpacker.c
+++ b/ext/msgpack/unpacker.c
@@ -245,7 +245,7 @@ static inline int _msgpack_unpacker_stack_push(msgpack_unpacker_t* uk, enum stac
     return PRIMITIVE_CONTAINER_START;
 }
 
-static inline VALUE msgpack_unpacker_stack_pop(msgpack_unpacker_t* uk)
+static inline size_t msgpack_unpacker_stack_pop(msgpack_unpacker_t* uk)
 {
     return --uk->stack.depth;
 }
--- a/ext/msgpack/unpacker.c
+++ b/ext/msgpack/unpacker.c
@@ -79,12 +79,9 @@ void msgpack_unpacker_static_destroy(void)
 
 #define HEAD_BYTE_REQUIRED 0xc1
 
-static inline msgpack_unpacker_stack_t* _msgpack_unpacker_new_stack(void) {
-    msgpack_unpacker_stack_t *stack = ZALLOC(msgpack_unpacker_stack_t);
+static inline void _msgpack_unpacker_stack_init(msgpack_unpacker_stack_t *stack) {
     stack->capacity = MSGPACK_UNPACKER_STACK_CAPACITY;
     stack->data = msgpack_rmem_alloc(&s_stack_rmem);
-    /*memset(uk->stack, 0, MSGPACK_UNPACKER_STACK_CAPACITY);*/
-    return stack;
 }
 
 void _msgpack_unpacker_init(msgpack_unpacker_t* uk)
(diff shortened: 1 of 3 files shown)


ext/msgpack/unpacker.c
```diff
@@ -343,14 +335,10 @@ static inline int read_raw_body_begin(msgpack_unpacker_t* uk, int raw_type)
             reset_head_byte(uk);
             uk->reading_raw_remaining = 0;
 
-            msgpack_unpacker_stack_t* child_stack = _msgpack_unpacker_new_stack();
-            child_stack->parent = uk->stack;
-            uk->stack = child_stack;
-
+            _msgpack_unpacker_stack_push(uk, STACK_TYPE_RECURSIVE, 1, Qnil);
```
peterzhu2118, line 338: Since a child no longer creating a new stack, does this change mean that recursive types will trigger a stack overflow much sooner than before?

ext/msgpack/unpacker.c
```diff
@@ -232,35 +224,35 @@ static inline int object_complete_ext(msgpack_unpacker_t* uk, int ext_type, VALU
...
 
-    uk->stack->depth++;
+    uk->stack.depth++;
     return PRIMITIVE_CONTAINER_START;
 }
 
 static inline VALUE msgpack_unpacker_stack_pop(msgpack_unpacker_t* uk)
 {
-    return --uk->stack->depth;
+    return --uk->stack.depth;
```
peterzhu2118, line 250: I don't think this function is returning a VALUE. `depth` is a `size_t`.

ext/msgpack/unpacker.c
```diff
@@ -343,14 +335,10 @@ static inline int read_raw_body_begin(msgpack_unpacker_t* uk, int raw_type)
...
-            msgpack_unpacker_stack_t* child_stack = _msgpack_unpacker_new_stack();
-            child_stack->parent = uk->stack;
-            uk->stack = child_stack;
-
+            _msgpack_unpacker_stack_push(uk, STACK_TYPE_RECURSIVE, 1, Qnil);
             int raised;
             obj = protected_proc_call(proc, 1, &uk->self, &raised);
-            uk->stack = child_stack->parent;
-            _msgpack_unpacker_free_stack(child_stack);
+            msgpack_unpacker_stack_pop(uk);
```
peterzhu2118, line 341: Can we assert that the pop here is actually popping a `STACK_TYPE_RECURSIVE` to ensure that we don't have a stack inconsistency issue?

ext/msgpack/unpacker.c
byroot: Possibly, yes. But the stack size if 128, that's quite a deep document, I don't think that's something you are likely to hit..

That said, your comment make me realize there's just no check whatsoever for the stack size :/

Would be worth adding one.

ext/msgpack/unpacker.c
byroot: Good catch, seems like it has been like this for long, I'll fix it.

ext/msgpack/unpacker.c
byroot: We don't quite have a good setup for asserts in the gem, but yes that would be valuable to put something like that in place.

ext/msgpack/unpacker.c
byroot: Nevermind, there's a check for it:

```c
    if(uk->stack.capacity - uk->stack.depth <= 0) {
        return PRIMITIVE_STACK_TOO_DEEP;
    }
```

peterzhu2118 (approved):

PR #373, opened 2024-11-11
Reorder msgpack_packer struct

byroot:
Reduce its size from 224 to 216B.

### DIFF
--- a/ext/msgpack/packer.h
+++ b/ext/msgpack/packer.h
@@ -31,14 +31,14 @@ typedef struct msgpack_packer_t msgpack_packer_t;
 struct msgpack_packer_t {
     msgpack_buffer_t buffer;
 
-    bool compatibility_mode;
-    bool has_bigint_ext_type;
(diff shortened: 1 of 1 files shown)


ext/msgpack/packer.h
```diff
@@ -31,15 +31,15 @@ typedef struct msgpack_packer_t msgpack_packer_t;
...
     VALUE to_msgpack_arg;
 
     VALUE buffer_ref;
 
+    bool compatibility_mode;
+    bool has_bigint_ext_type;
+    bool has_symbol_ext_type;
+
     /* options */
     bool comaptibility_mode;
```
peterzhu2118, line 44: Btw, if you use a bitfield, you can even turn this from 16B down to 8B

byroot: I'm never too sure how bitfields impact performance. Probably negligible, but I only made this change because it was an obvious small waste.

PR #379, opened 2024-11-14
Various micro optimizations

casperisfine:
Nothing major, mostly use more direct APIs to bypass safety checks that we can assert aren't needed.

e.g. `OBJ_FROZEN_RAW` doesn't bother checking if the VALUE is an immediate, which we know it isn't because we know it's a T_STRING.

Or `ENCODING_GET_INLINED` doesn't check if we're dealing with an extended encoding, which in our case we don't need to know.


casperisfine:
Alright, it passed on the retry, so I'm gonna assume it's unrelated. Logs are here: https://github.com/msgpack/msgpack-ruby/actions/runs/11837028381/job/32983173005

PR #387, opened 2025-12-11
[PROTOTYPE] Add optimized_struct option for fast Struct serialization

gmalette:
This is a prototype, please don't actually merge this as I cannot write C code and this is all Claude-generated.

The idea is to add a C-level fast path for serializing and deserializing Ruby Struct types, bypassing Ruby proc callbacks entirely.

```
  factory.register_type(0x01, MyStruct, optimized_struct: true)
```

The wire format is identical to the recursive proc-based approach.

### DIFF
--- a/bench/various_appraches_bench.rb
+++ b/bench/various_appraches_bench.rb
@@ -1,0 +1,770 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "bundler/inline"
+
+gemfile(true) do
+  source "https://rubygems.org"
+  if ENV["MSGPACK_PATH"]
+    gem "msgpack", path: ENV["MSGPACK_PATH"]
+  else
+    gem "msgpack", git: "https://github.com/Shopify/msgpack-ruby.git", branch: "gm/ref-tracking"
+  end
+  gem "benchmark-ips"
+end
+
+require "msgpack"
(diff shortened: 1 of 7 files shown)


bench/various_appraches_bench.rb
```diff
@@ -0,0 +1,770 @@
...
+    num_variants: scenario[:variants],
+    num_selling_plan_groups: scenario[:spg],
+    num_selling_plans_per_group: scenario[:sp],
+    num_product_metafields: scenario[:product_metafields] || 0,
+    num_variant_metafields: scenario[:variant_metafields] || 0,
+  )
+
+  puts "\nBenchmarking scenario: P:#{scenario[:products]} V:#{scenario[:variants]} PM:#{scenario[:product_metafields]} VM:#{scenario[:variant_metafields]} SPG:#{scenario[:spg]} SP:#{scenario[:sp]}"
+
+  payloads = coders.map { |coder| coder.factory.dump(data) }
```
byroot, line 677:
```suggestion
  payloads = coders.map { |coder| coder.factory.dump(data).freeze }
```

FYI: If the payload is frozen, it saves a bunch of work for msgpack. The diff isn't massive, but on you simple benchmark it gives another ~0.10x  over marshal.

